PR #8, opened 2019-02-14
Remove haste imports from Slider 

pbitkowski:
removed haste imports from Slider component


michalchudziak:
Nice! How about other imports like `'ImageSource'`?

pbitkowski:
Flow types fixed :)

PR #56, opened 2019-04-26
Use AndroidX support lib

aarongrider:
Summary:
---------

Use AndroidX in place of legacy Android support libraries. This is required for any Android projects that use AndroidX for support libraries.

See: https://developer.android.com/jetpack/androidx/migrate

Specifically react-native-gesture-handler currently has a 4 PRs to migrate to AndroidX being blocked by this down stream dependency for our tests that require react-native-slider.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -37,6 +37,6 @@ repositories {
 
 dependencies {
   //noinspection GradleDynamicVersion
-  api "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
+  api "androidx.appcompat:appcompat:${rootProject.ext.supportLibVersion}"
   api 'com.facebook.react:react-native:+'
 }
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -2,3 +2,6 @@ ReactNativeSlider_compileSdkVersion=28
 ReactNativeSlider_buildToolsVersion=28.0.3
 ReactNativeSlider_targetSdkVersion=27
 ReactNativeSlider_minSdkVersion=16
+
+android.enableJetifier=true
+android.useAndroidX=true
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -1,6 +1,6 @@
 ReactNativeSlider_compileSdkVersion=28
 ReactNativeSlider_buildToolsVersion=28.0.3
-ReactNativeSlider_targetSdkVersion=27
+ReactNativeSlider_targetSdkVersion=28
 ReactNativeSlider_minSdkVersion=16
 
 android.enableJetifier=true
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -135,7 +135,7 @@ android {
 
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
-    implementation "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
+    implementation "androidx.appcompat:appcompat:${rootProject.ext.supportLibVersion}"
     implementation "com.facebook.react:react-native:+"  // From node_modules
     implementation project(':react-native-slider')
 }
--- a/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -8,7 +8,7 @@ package com.reactnativecommunity.slider;
 
 import android.content.Context;
 import android.os.Build;
-import android.support.v7.widget.AppCompatSeekBar;
+import androidx.appcompat.app.AppCompatSeekBar;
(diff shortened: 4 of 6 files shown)


android/build.gradle
```diff
@@ -37,6 +37,6 @@ repositories {
 
 dependencies {
   //noinspection GradleDynamicVersion
-  api "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
+  api "androidx.appcompat:appcompat:${rootProject.ext.supportLibVersion}"
```
dulmandakh: you can remove this because AndroidX already bundled in RN

android/build.gradle
```diff
@@ -37,6 +37,6 @@ repositories {
...
   api 'com.facebook.react:react-native:+'
```
dulmandakh: this must be implementation, not api

example/android/app/build.gradle
```diff
@@ -135,7 +135,7 @@ android {
 
 dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
-    implementation "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
+    implementation "androidx.appcompat:appcompat:${rootProject.ext.supportLibVersion}"
```
dulmandakh: remove this line, because AndroidX is bundled in RN.

android/gradle.properties
```diff
@@ -1,4 +1,7 @@
 ReactNativeSlider_compileSdkVersion=28
 ReactNativeSlider_buildToolsVersion=28.0.3
-ReactNativeSlider_targetSdkVersion=27
-ReactNativeSlider_minSdkVersion=16
\ No newline at end of file
+ReactNativeSlider_targetSdkVersion=28
+ReactNativeSlider_minSdkVersion=16
+
+android.enableJetifier=true
```
dulmandakh: maybe you can remove it, please check if it works.

android/gradle.properties
```diff
@@ -1,4 +1,7 @@
...
+android.useAndroidX=true
```
dulmandakh: maybe you can remove it, please check if it works.

android/build.gradle
Salakar: @dulmandakh isn't AndroidX only for RN 0.60+ though? Making removing this backwards incompatible for older RN versions?

dulmandakh: IIRC, **appcompat** from Android Support Library is bundled in React Native since 0.59. So 0.60 will included from AndroidX too. Thus, it's not required to have appcompat as dependency as it'll add extra work.

dulmandakh: it's included in 0.58.

michalchudziak:
@aarongrider Can you please check the comments above? I'll be glad to merge your PR after you resolve the discussion.

android/build.gradle
aarongrider: I see 'javax.inject:javax.inject:1' and 'com.android.support:appcompat-v7:28.0.0' bundled into RN 58.4, but not AndroidX. I think we do have to include it here and in the example project?

aarongrider: We also may want to just hold on this PR until 60 comes out?

aarongrider:
Happy to try and resolve, but I'm wondering if we should wait to merge this until RN 60 is released with AndroidX support. As far as I understand, AndroidX projects are incompatible with legacy support projects when compiled from source. This library shouldn't be a huge issue for consumers since you distribute compiled aar's, but it's something we should keep in mind.

dulmandakh:
How about splitting it into two PRs, one for Gradle cleanup and AndroidX.

michalchudziak:
+1 for what @dulmandakh said. I'd appreciate grade cleanup in this PR, leaving AndroidX support to 0.60 release.

Btw. I've just merged detox PR which conflicts with this PR. sorry about that. Thanks for all the work @aarongrider

aarongrider:
Sounds good! Will do.

Taylor123:
seeing this issue today #82 after these releases: https://developers.google.com/android/guides/releases

is there any movement on this? There are many libraries in the RN community that are affected since they use old support dependencies

natasha08n:
I have the same question, the app doesn't work, the problem is pretty similar to the issue [#82](https://github.com/react-native-community/react-native-slider/issues/82)

michalchudziak:
@Taylor123 @natasha08n I'm working on 0.60 update, you can expect rc version soon

michalchudziak:
Sorry @aarongrider, I'm deeply sorry, but I have to close this PR. I've tried to use your changes in 0.60 update PR, but there were too many conflicting files since I had to re-create example from the scratch. Also, most of these changes aren't suitable for a final 0.60 release, please refer to the https://github.com/react-native-community/react-native-slider/pull/87 for AndroidX support.

PR #66, opened 2019-05-13
Add onSlidingStart prop on SliderProps interface in index.d.ts

fossamagna:
Summary:
---------

I use TypeScript to develop my react-native app. When I use `onSlidingStart` prop, I received a error message from transpiler, because of `onSlidingStart` prop definition is not contains in `typings/index.d.ts`.


michalchudziak:
Thanks a lot, I've missed that.

michalchudziak:
Released in 1.1.2

PR #70, opened 2019-05-16
Updated Readme

SkyTreasure:
Updated manual linking.

Summary:
---------

Test Plan:
----------


michalchudziak:
Not sure if `@` is a valid character for gradle. Even if it is, it's important to keep `android/settings.gradle` and `android/app/build.gradle` in sync. The current manual seems right, if you had something else in your mind, please write the comment below.

PR #87, opened 2019-06-30
React Native 0.60 support

michalchudziak:
Summary:
---------

Updated an example to use react native 0.60.0-rc.3, updated the module to work with following RC, updated the dependencies to the latest versions.

This should be an RC release as soon as it gets merged.

Disabled `detox` tests until full 0.60 support lands in place. Tracking issue: https://github.com/react-native-community/react-native-slider/issues/88

Test Plan:
----------

Run the example app and test if everything works correctly.


.gitignore
```diff
@@ -54,4 +54,6 @@ buck-out/
 
 # Bundle artifact
 *.jsbundle
-.tmp/
\ No newline at end of file
+.tmp/
+
+Pods/
```
thymikee: shouldn't this be `ios/Pods/`?
```suggestion
ios/Pods/
```

js/Slider.js
```diff
@@ -249,22 +249,16 @@ const Slider = (
       ref={forwardedRef}
       style={style}
       onChange={onChangeEvent}
-      onSlidingStart={onSlidingStartEvent}
-      onSlidingComplete={onSlidingCompleteEvent}
-      onValueChange={onValueChangeEvent}
+      onRNCSliderSlidingStart={onSlidingStartEvent}
+      onRNCSliderSlidingComplete={onSlidingCompleteEvent}
+      onRNCSliderValueChange={onValueChangeEvent}
```
thymikee: can't we avoid changing public API?

.gitignore
michalchudziak: I think it works recursively

js/Slider.js
michalchudziak: I haven't changed a public API, we still expose `onSlidingStartEvent `,  `onSlidingCompleteEvent` and `onValueChangeEvent` and use this class to change props names under the hood. W/o this change the app throws an error (built by xcode)

thymikee: Oh, that's great!

thymikee:
FYI, let's wait for this to ship, should be in the next RC: https://github.com/react-native-community/cli/pull/451

paolospag:
I tried to use `react-native-slider` with `react-native@0.60.0-rc.3`, but I encountered the following error on iOS:

Simulator Screen Shot - iPhone 8 - 2019-07-02 at 18 04 05

Is this related by the lack of support for React Native ^0.60?

thymikee:
Yes, this is fixed in this PR

PR #117, opened 2019-10-17
feat(slider): adds support for touch to seek to iOS slider

shwetsolanki:
Summary:
---------

iOS Slider doesn't have tap to seek feature enabled by default. In this PR, I have added the support to iOS to have parity with Android

Test Plan:
----------

Tapping on the track bar seeks to that position.


wojteg1337:
Hello @shwetsolanki,

Good catch, will check it later today/tomorrow

michalchudziak:
@shwetsolanki We'd like to merge this PR, could you fix conflicts?

shwetsolanki:
@michalchudziak done

shwetsolanki:
@michalchudziak are we good with the PR?

cristianoccazinsp:
Is this live on 2.0.5? I can't seem to make it work. Tapping on anywhere on the slider will always move the button to the beginning. IOS 13.2.2

SiyabongaNzulwana:
On iOS the slider does not work if I dragg it, works fine on android. please help

```
          <Slider
            maximumTrackTintColor='#FFF'
            minimumTrackTintColor='#FFF'
            thumbTintColor='#FFF'
            value={currentTime / duration} // slier input is 0 - 1 only so we want to convert sec to 0 - 1
            onValueChange={this.onslide}
            disabled={false}
            onSlidingStart={()=> this.start}
          />
```

SiyabongaNzulwana:
@shwetsolanki please help, I am using `    "@react-native-community/slider": "^1.1.2",
` and the feat(slider): adds support for touch to seek to iOS slider #117 does not work. I am on react native `59` please advise

shwetsolanki:
Will check and revert soon

src/ios/RNCSlider.m
```diff
@@ -10,6 +10,22 @@
...
+    if (self) {
+        tapGesturer = [[UITapGestureRecognizer alloc] initWithTarget: self action:@selector(tapHandler:)];
+        [tapGesturer setNumberOfTapsRequired: 1];
+        [self addGestureRecognizer:tapGesturer];
+    }
+    return self;
+}
+- (void)tapHandler:(UITapGestureRecognizer *)gesture {
+    CGPoint touchPoint = [gesture locationInView:self];
+    [self setValue:touchPoint.x / self.bounds.size.width animated: YES];
```
mikehardy, line 28: @shwetsolanki  / @michalchudziak  this was so close to correct - I wanted tap to move the slider on android and ios so I dug in here because it was reverted in #136 - it looks to me like this is generating the "percent of the slider that was tapped on" but the manager method that handles the value-set is expecting "percent of slider scaled up to maximumValue" and indeed this works when I test it:

```objective-c
    [self setValue:self.maximumValue * touchPoint.x / self.bounds.size.width animated: YES];
```

I'll post a fresh PR that implements it this way with this commit picked as the start of it

PR #120, opened 2019-10-24
quick fix for 114. Evaluates getThumb before accesing it for setting …

fcopardo:
…setting the thumb tint color

Summary:
---------

Hi! currently, the setThumbTintColor fails on android builds in release mode because the thumb hasn't been set. The change makes the call safe, thus avoiding crashes in production mode when sliders are added to hidden parts of the hierarchy.

Test Plan:
----------

Run a test app in release mode. Crashes due to com.facebook.react.bridge.JSApplicationIllegalArgumentException: Error while updating property 'thumbTintColor' of a view managed by: RNCSlider - Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.graphics.drawable.Drawable.setColorFilter(int, android.graphics.PorterDuff$Mode)' on a null object reference should not happen.


android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -169,10 +169,12 @@ public void setStep(ReactSlider view, double value) {
 
   @ReactProp(name = "thumbTintColor", customType = "Color")
   public void setThumbTintColor(ReactSlider view, Integer color) {
-    if (color == null) {
-      view.getThumb().clearColorFilter();
-    } else {
-      view.getThumb().setColorFilter(color, PorterDuff.Mode.SRC_IN);
+    if(view.getThumb()!=null){
```
wojteg1337: add space before bracket

wojteg1337: fix formatting

wojteg1337: fix formatting

wojteg1337:
Hello @fcopardo, 
could you please resolve conflicts and fix linting so I can merge and release fix?

PR #125, opened 2019-11-14
Fix missing thumbImage on android release mode

pheis:
## Summary
Fixes https://github.com/react-native-community/react-native-slider/issues/119

In debug mode RNs "Image.resolveAssetSource" method returns http-uri and in release mode that returns android resource name. Previously the case where it returns an android resource name was not handled.


wojteg1337:
@pheis thanks for the contribution, great work

PR #164, opened 2020-03-30
Enhanced Customization

ShaMan123:
Summary:
---------
### Customization of the slider's appearance.
I've added the option to completely control the appearance of `thumb`, `minimumTrack`, `maximumTrack` by introducing them as props accepting one of [`FunctionComponent`, `Element`,  `RefObject`, `viewTag`]. 
Once rendered they are drawn onto the slider. If they change and redraw this propagates to the slider as well.
**This makes drawing a linear gradient as the track or the thumb and achieving something like a color picker extra simple**.

### Behind the Scenes
`thumb`, `minimumTrack`, `maximumTrack` are first rendered if necessary (in an invisible container), then their `viewTag` is passed to native for drawing. In native I've created a set of classes handling `drawable` creation and updating.
Updating is done via 2 ways:
1. native `onDrawListener` - once the view redraws, slider is informed
1. props changed informers - once a view receives new props in native, slider is informed (important for animations, hardware rendering...)

With these changes `Slider` is a lot more reactive and customizable.
I believe iOS will be able to support this as well, unfortunately I don't have a mac so I can't work on it. Hoping someone will take it on.

Test Plan:
----------
Added example and updated `jest`

slider show


ShaMan123:
My fork was stale... so I've committed some adjustments... I should have updated it before I started the work on it.

ShaMan123:
OK this seems stable.
The only thing to be done is to make the track height adjustable.

ShaMan123:
Added support for animations as well.
**Caveat**:
In order to use `transform` styling you need to wrap the view with another view.
This can be solved as well by adding logic to `DrawableHandler` class

next

ShaMan123:
android is done apart from mentioned caveat:

> Added support for animations as well, see **EDITED** first comment
> **Caveat**:
> In order to use `transform` styling you need to wrap the view with another view.
> This can be solved as well by adding logic to `DrawableHandler` class.
> 
> next

wojteg1337:
Hello, nice to see that big input from contributor, I need a while to review that code :)

ShaMan123:
Great. I am experimenting with more optimizations, so I might commit more changes.

ShaMan123:
The last push fixes the mentioned caveat, so now transforms are supported on top level as well.

> **Caveat**:
> In order to use `transform` styling you need to wrap the view with another view.
> This can be solved as well by adding logic to `DrawableHandler` class.
>

ShaMan123:
Committed a bunch of fixes.
The biggest would be making `maximumTrack` a part of the progress drawable as opposed to the background drawable. This was causing trouble when applying transformations because of it's size being the full size of the slider.
Drawable hierarcy:
```bash
 main progress drawable (slider.getProgressDrawable())
├── background drawable (`android.R.id.background`) -> backgroundTrack
└── progress drawable (`android.R.id.progress`)
   ├── progress drawable (`android.R.id.secondaryProgress`) -> maximumTrack
   └── progress drawable (`android.R.id.progress`) -> minimumTrack
```

ShaMan123:
This PR is in a good condition.

### Next Up
I have explored an approach that is extremely performant and can run on 60fps but requires more work.
This will be done with `ReactDrawableGroup` which is a matching drawable hierarchy to the view it's drawing. This is done to enable animating a child view in the tree with `Animated` or redrawing it because of invalidation.
This already works with minimum/maximum tracks. I was struggling to get it work with the thumb.
#### Drawbacks
The overhead in this approach is creating the tree, so it should be done as less as possible (I am aiming for once on initialization). Then it should listen to changes made to the hierarchy (drawing, add/remove child) and handle the specific layer that was changed.
#### Current Approach
The current approach draws a `BitmapDrawable` for the entire hierarchy (which is relatively fast and lightweight), listens to changes and redraws for every change (except if the top most view receives animations - this is possible due to achievements made from the **NEXT** approach).
#### Summing Up
Mixing the two is very bad because you loose from each, especially from recreating the drawable hierarchy. This is why I stabilized the PR on the initial approach and from now on will commit changes to a different branch (if I decide to continue working on this... ;)).

### Next Approach TODO
1. Once a view draws it's matching drawable will redraw and only it.
1. Add/remove view from the hierarchy.
1. Thumb drawable

wojteg1337:
I set up the meeting with other developers to review the code and see what is there because thats really a lot of code

michalchudziak:
Hey @ShaMan123 ! Thanks for such an important and comprehensive PR. We are happy and thankful to have this functionality, please let us know once the PR is complete and ready for review. I'll start looking onto the code already, but would like to perform an end to end test once it's ready.

michalchudziak:
I would love to have corresponding iOS implementation. Looking for volunteers <3 :)

ShaMan123:
@michalchudziak I'm happy to hear you understand the importance!
I'll let you know.

ShaMan123:
@michalchudziak I have found a better solution and created a new PR for it. So I'm closing this one.
This approach is too complex and has lots of pitfalls and caveats.

PR #173, opened 2020-04-12
ENHANCED CUSTOMIZATION

ShaMan123:
Continuing #164 

This approach is much simpler than #164 .
It consists upon a `ViewGroup` laying out it's children above/below the slider. It's great. No caveats or performance hits.
 

### Structure
Rendering a slider component now renders a view with 5 children:
1. backgroundTrack
1. maximumTrack
1. minimumTrack
1. slider
1. thumb

 The custom views are translated and scaled to appear exactly in position. This is done to avoid performance degradation, and so it does. Besides that I've taken pieces of valid code from the previous PR that fixes stuff.

#### Ready for reviewing!

This approach is much more agnostic, so writing the iOS implementation can be pretty much the same.


BartoszKlonowski:
Hello @ShaMan123!
Thank you for taking the time to implement such a huge change and sorry for that much delayed response...

I briefly went through the implementation and I would like to "resurrect" this topic.
Are you planning to, or would you like, to implement the missing iOS support?
I can spend some time this weekend (29.04 - 01.05) to rebase this and resolve all the conflicts.

Just in case: Feel no pressure, I already appreciate the work. Let me just know please what are your thoughts.

ShaMan123:
@BartoszKlonowski I no longer work on RN projects. Anyways I don't have a mac. So my answer is NO.

PR #175, opened 2020-04-28
Add accessibility improvements for screen readers

orofbrown:
Summary:
---------
Add accessibility improvements for screen readers to announce custom units.

Currently there are no capabilities for custom screen reader output with this component. This PR would allow consumers to specify a custom unit to be read by the screen reader, instead of just the default `percent` value. It is an optional field, so if nothing is specified it will use the default behavior.
There is one minor caveat:
- On iOS everything is overridden as expected
- On Android, the default values are still announced, along with the additional custom values. This seems to be due to some limitations of accessibility in RN 61, and the `ReactAccessibilityDelegate`

Test Plan:
----------
I tested this by replacing the yarn library with my local copy in my existing application, that I work on full-time. Doing the same with an example application should work fine, but this requires having a physical device available. The changes should be
- On accessibility focus, the screen reader announces the current value of the slider with the custom units that were specified
- On value update, the screen reader announces the updated value of the slider with the custom units that were specified

### DIFF
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -11,18 +11,18 @@ import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.drawable.BitmapDrawable;
 import android.os.Build;
-
-import androidx.annotation.RequiresApi;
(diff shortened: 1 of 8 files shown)


michalchudziak:
Thanks @orofbrown! I'll add it to `3.0.0` release :)

marcshilling:
@orofbrown these props don't seem to work at all for me 🤷

PR #178, opened 2020-05-04
Update value on web onTouchEnd so clicking also updates the slider position

piemonkey:
Summary:
---------

I had an issue on web at work. We ended up not using a different approach but I spotted the difference between web and mobile behaviour and how to fix it before we did.

Test Plan:
----------

I used the updated version in an ejected Expo app as it was the easiest way to get react-native-web support for free. Test by clicking on the slider instead of dragging.

### DIFF
--- a/src/js/RNCSliderNativeComponent.web.js
+++ b/src/js/RNCSliderNativeComponent.web.js
@@ -268,24 +268,30 @@ const RCTSliderWebComponent = React.forwardRef(
         if (value !== withinBounds) {
           setValue(withinBounds);
           onValueChange(withinBounds);
+          return withinBounds;
         }
+        return hardRounded;
       },
       [minimumValue, maximumValue, value, onValueChange],
     );
 
-    const onTouchEnd = () => {
-      onSlidingComplete(value);
-    };
-
-    const onMove = event => {
-      const {locationX: x} = event.nativeEvent;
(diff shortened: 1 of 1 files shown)


Sharcoux:
I was wondering about this. I really hesitated, my concern was that we could change the value by mistake while scrolling through the page... What do you think?

piemonkey:
Hmm, interesting point. It'd only be a problem on touch screens but I don't know of a way in react-native-web to tell touches from clicks to filter them out. This behaviour mimics the native components though so not sure how they prevent the same problem. Do they prevent the same problem?

EvanBacon:
Related https://github.com/necolas/react-native-web/issues/1568

michalchudziak:
Hey, I'd like to have this PR in, and release it after implementing https://github.com/react-native-community/react-native-slider/issues/211

michalchudziak:
@piemonkey do you mind resolving conflicts?

michalchudziak:
> Hmm, interesting point. It'd only be a problem on touch screens but I don't know of a way in react-native-web to tell touches from clicks to filter them out. This behaviour mimics the native components though so not sure how they prevent the same problem. Do they prevent the same problem?

They do not at this point, but we are aiming to fix that :)

piemonkey:
@michalchudziak, Sure, I should be able to take a look this week.

PR #136, opened 2019-11-26
Fix touch to seek flicker on iOS

michalchudziak:
Summary:
---------

Fixes #134 #131

Test Plan:
----------

Touch to seek and confirm that there's no flicker.

Issue #134, opened 2019-11-22
[iOS] Touch on slider bar brings the thumb to beginning of the bar


ludovic-gonthier:
## Environment

```
System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
    Memory: 1.36 GB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.11.0 - ~/.nvm/versions/node/v10.11.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.11.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 12.4, macOS 10.14, tvOS 12.4, watchOS 5.3
    Android SDK:
      API Levels: 28
      Build Tools: 28.0.3
      System Images: android-28 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
    Xcode: 10.3/10G8 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0
    react-native: 0.61.1 => 0.61.1
```

## Description

On iOS, when I try to touch the slide to go increase/decrease its value, the thumb always goes to the beginning of the slider without triggering the `onValueChange` callback.
On Android, this behavior works fine.

## Reproducible Demo

```jsx
<Slider
  minimumValue={6}
  value={16}
  maximumValue={26}
  onValueChange={(value) => console.log('Value changed: ', value}
  step={1}
/>
```

mcousillas6:
I'm having the same issue, is there any way to disable the tap to seek until this is resolved?

BunlongOung:
@mcousillas6 You can downgrade to use `2.0.3` for now.

BunlongOung:
I'm facing this issue, too. Downgrade to 2.0.3 can solve the problem for now since it does not handle touch on slide bar

michalchudziak:
I guess it'd be best if we add an option to toggle that behavior via props. Anyone interested in sending a PR? cc @wojteg1337

mcousillas6:
> I guess it'd be best if we add an option to toggle that behavior via props. Anyone interested in sending a PR? cc @wojteg1337

[...]

Issue #131, opened 2019-11-19
Slider still jumping all over the place when sliding back and forth using step={1}


schumannd:
This used to be an issue #109 for users using old react native versions. Now that we updated, the issue still seems to persist.

Using the latest version:

    "@react-native-community/slider": "2.0.3",
    "react-native": "0.61.4",

Krizzu:
@schumannd What iOS version have you tested?

schumannd:
iOS 13.1.2

michalchudziak:
This issue should be fixed in version `2.0.6`. Please check it out and re-open this if it still occurs :)

schumannd:
It was an issue caused by my handling of onValueChange to begin with. Thanks for your help!

joshbg2k:
> It was an issue caused by my handling of onValueChange to begin with. So it probably wasn't broken in the end. Thanks for your help!

@schumannd How did you handle it? I have the same issue but with RN 0.55.4.

schumannd:
@joshbg2k actually we didn't.

The issue arises for us, if we add any state update to the `onValueChange` vs. `currentValue` loop. Removing this, or only using `onSlidingComplete` avoids the problem. For us this was not an option though, as we want a "speech bubble" to always stay with the sliding head. Is this situation similar to yours?

joshbg2k:
@schumannd Yeah it's really similar. My slider enables a range of values in steps, so when the user slides in either direction, the label for each step changes based on it being enabled or disabled, so I use `onValueChange` to update the state and manage that. So I think we're both in the same boat - we want richer user feedback _while_ the user is using the slider. Looks like I'll have to use `onSlidingComplete` and update multiple labels at once at the end of the interaction. Not ideal but at least the feature will be usable until a fix or alternative is available.

schumannd:
@joshbg2k, @michalchudziak I would say it is not fixed then. It used to work before we updated to RN 61.4. So it might be related to a change in how RN updates state. I also couldn't find any reference to `currentValue` in the README. Is that not supported anymore? Maybe that's where the issue lies.

lkoehl:
For me the slider also jumps around, but also when step > 1. It doesn't change if I don't set the state in ```onValueChange```. For right now I needed to switch to another slider library. 
Also I am using the latest version ```2.0.8```.

Jason-Cooke:
@michalchudziak Could we reopen this please?

mikehardy:
[...]

PR #201, opened 2020-07-03
Re add ios tap handler

mikehardy:
Summary:
---------

First thanks for this package everyone, I appreciate it

I want the slider on iOS to handle tap events by sliding to the tap.

It's been tried before: #117 implemented tap-handling for iOS, and it worked, mostly

But: #131 #134 were documented regressions from #117

So #136 reverted the change from #117 as the understandable quick fix

I investigated the original PR because I really wanted it to work, and discovered that it was just a mis-match between the values the tap handler propagated compared with what the value consumer wanted

Specifically:
- value emitted was pure 0 through 1 inclusive percentage where the slider was tapped
- value consumed was expected to be the range value from minimum to maximum

So:
- I reverted the revert from #136 / @michalchudziak 
- That brings us back to the #117 state where we have tap-handling from @shwetsolanki but with a bug
- I altered the math in the tap handler to take the original raw slider percentage and emit instead the range value

Now it works?

Test Plan:
----------

1. First just hacking around I tested this in my project directly with a node_modules edit, it worked! For positive min/max and a 1 step value it was perfect
1. Then I forked and cloned and checked it in the example - it didn't work! 
1. So I added onSlideComplete handling in the example and now it worked everywhere except negative min value
1. So I corrected the math to handle negatives (similar to existing code)

Now it works everywhere I believe.
Interested parties that should test if/when this is integrated are @schumannd @ludovic-gonthier and @mcousillas6  and @BunlongOung

I will attach a patch-package file in a follow-on comment that you can drop in patch-package for use with the current published version in order to easily verify. I would love any success (or error) reports

Cheers

### DIFF
--- a/example/SliderExample.js
+++ b/example/SliderExample.js
@@ -31,6 +31,7 @@ class SliderExample extends React.Component<$FlowFixMeProps, $FlowFixMeState> {
         <Slider
           {...this.props}
           onValueChange={value => this.setState({value: value})}
+          onSlidingComplete={value => this.setState({value: value})}
         />
       </View>
     );
--- a/src/ios/RNCSlider.m
+++ b/src/ios/RNCSlider.m
@@ -23,9 +23,12 @@
     }
     return self;
 }
+
 - (void)tapHandler:(UITapGestureRecognizer *)gesture {
     CGPoint touchPoint = [gesture locationInView:self];
-    [self setValue:touchPoint.x / self.bounds.size.width animated: YES];
+    float rangeWidth = self.maximumValue - self.minimumValue;
+    float sliderPercent = touchPoint.x / self.bounds.size.width;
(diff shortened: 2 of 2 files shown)


mikehardy:
Use [patch-package](https://github.com/ds300/patch-package) in your project and drop this in your patches folder (without the .txt ending I needed for github) to test - 🙏 please someone report success or failure, ideally someone that had problems before

Thanks!

[@react-native-community+slider+3.0.0.patch.txt](https://github.com/react-native-community/react-native-slider/files/4871484/%40react-native-community%2Bslider%2B3.0.0.patch.txt)

mikehardy:
@brentvatne / @michalchudziak any interest? has received more wide spread testing in my work project, no issues so far, seems to address all the existing issues with the previous attempt...

PR #210, opened 2020-07-17
Add Windows Slider.

rectified95:
Summary: Adding the Windows implementation of this module.
---------
Implemented the following properties:
  - maximumValue, minimumValue, value, step
  - inverted, disabled, vertical (Windows-specific)
  - minimumTrackTintColor, maximumTrackTintColor
  - thumbTintColor, thumbImage
  - onValueChange, onSlidingStart, onSlidingComplete

Closes microsoft/react-native-windows#4612 

Updated the README with information on how to run the test app on Windows, and how to link this module to existing projects.
(Also moved the example app's one into the main file in order to improve coherence).

Updated the version of React Native for this module to >=62 - I'd like to begin a discussion on whether this is fine for this module, or should be done only for Windows, which requires that version.

Also, spotted dead code in `Slider.js` - the default prop values don't get picked up at all - changing them doesn't alter the app's behavior. Given this and the fact that a default value of 0 for the `step` prop seems mistaken, I'm also updating it by placing it in the component definition (can be overridden easily). (defaultProps is not available on components returned by forwardRef, as is the case here - https://github.com/facebook/flow/issues/7467)

Test Plan:
----------
As a subsequent PR, I'd like to add a basic CI loop for this. Since there are issues with Windows builds on CircleCI, it'd have to be through the addition of Github actions.

You can run the test app on Windows to see it in action. 

 sliderWindows


README.md
```diff
@@ -308,17 +318,60 @@ To fix formatting errors, run the following:
...
+
+## Start the app
+
+- Run `yarn run:android` to run on Android
+- Run `yarn run:ios` to run on iOS
+- Run `yarn run:web` to run on web
+
+### Windows
+- Run `yarn run:windows` to run the bundler. Then, run the example app solution in Visual Studio.
+
```
asklar, line 341: can we simplify this? yarn run:windows should be able to do npx react-native run-windows?

README.md
```diff
@@ -308,17 +318,60 @@ To fix formatting errors, run the following:
...
+# Manual linking of the module on Windows
+On Windows, you need to manually link the `Slider` native module to your project.
+
+#### Add the SliderWindows project to your solution
+
+1. Open the solution in Visual Studio 2019.
+2. Right-click solution icon in Solution Explorer > Add > Existing Project.
+   Select 'D:\pathToYourApp\node_modules\@react-native-community\slider\windows\SliderWindows\SliderWindows.vcxproj'.
+
```
asklar, line 350: isn't this possible with the react-native CLI? most RN devs will be more comfortable with the CLI than VS, could we add instructions to do that?

example/SliderExample.js
```diff
@@ -171,6 +174,20 @@ export const examples = [
...
   },
+  {
+    title: 'Custom thumb (network image)',
+    platform: 'windows',
+    render(): Element<any> {
+      return (
+        <SliderExample
+          thumbImage={{
+            uri:
+              'https://cdn2.iconfinder.com/data/icons/minimalism/512/twitter.png',
```
asklar, line 185: maybe use an icon in the open domain as opposed to an icon with a license attached to it (plus this link is in a CDN so it may change?)

example/metro.config.windows.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = {
+  resolver: {
+    extraNodeModules: {
+      // Redirect react-native to react-native-windows
+      'react-native': rnwPath,
+      'react-native-windows': rnwPath,
+    },
+    // Include the macos platform in addition to the defaults because the fork includes macos, but doesn't declare it
+    platforms: ['ios', 'android', 'windesktop', 'windows', 'web', 'macos'],
```
asklar, line 27: what's windesktop? 🤔

example/package.json
```diff
@@ -5,16 +5,18 @@
   "scripts": {
     "start": "node ../node_modules/react-native/local-cli/cli.js start",
     "run:android": "react-native run-android",
+    "run:windows": "react-native start --use-react-native-windows --projectRoot ../",
```
asklar, line 8: why not react-native run-windows?

example/windows/SliderTestWindows/Package.appxmanifest
```diff
@@ -0,0 +1,49 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+
+<Package
+  xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10"
+  xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest"
+  xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10"
+  IgnorableNamespaces="uap mp">
+
+ <Identity
+    Name="359e3698-40e8-432a-a406-1213a9404462"
```
asklar, line 10: give the package a human readable name?

src/windows/SliderWindows/SliderView.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+        for (auto const& pair : propertyMap) {
+            auto const& propertyName = pair.first;
+            auto const& propertyValue = pair.second;
+
+            if (propertyName == "value") {
+                if (propertyValue.IsNull()) {
+                    this->ClearValue(xaml::Controls::Primitives::RangeBase::ValueProperty());
+                }
+                else {
```
asklar, line 72: is this using clang format? seems like the formatting is different than in our repo?

src/windows/SliderWindows/SliderView.cpp
```diff
@@ -0,0 +1,237 @@
...
+                    auto resDict = this->Resources();
+
+                    resDict.Insert(winrt::box_value(L"SliderThumbBackground"), propertyValue.To<winrt::Brush>());
+                    resDict.Insert(winrt::box_value(L"SliderThumbBackgroundPointerOver"), propertyValue.To<winrt::Brush>());
+                    resDict.Insert(winrt::box_value(L"SliderThumbBackgroundPressed"), propertyValue.To<winrt::Brush>());
+                }
+            }
+            else if (propertyName == "thumbImage") {
+                if (!propertyValue.IsNull()) {
+                    auto imgUriString = propertyValue.AsObject()["uri"].AsString();
```
asklar, line 158: is it guaranteed that the thumbimage is a url? i.e. it won't work with e.g. base64 encoded svg?

src/windows/SliderWindows/SliderWindows.vcxproj
```diff
@@ -0,0 +1,174 @@
...
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
```
asklar, line 59: we should only be using v142 (and not mixing; rnw uses 142)

example/windows/SliderTestWindows/App.cpp
```diff
@@ -0,0 +1,42 @@
...
+/// authored code executed, and as such is the logical equivalent of main() or
+/// WinMain().
+/// </summary>
+App::App() noexcept
+{
+    MainComponentName(L"example");
+
+#if BUNDLE
+    JavaScriptBundleFile(L"index.windows");
+    InstanceSettings().UseWebDebugger(false);
```
asklar, line 23: We removed this pattern at some point - not super recently but in the last few months - and we now use a MainPage component instead which is how the usual UWP app template works in VS, is this not using RNW 62+?

rectified95: I was using RNW62 and this is what got generated

README.md
rectified95: hmm, that might be a good idea. I just followed what we put in the docs

README.md
rectified95: You mean autolinking? I thought it's not available in RNW v62

example/SliderExample.js
rectified95: will do

src/windows/SliderWindows/SliderView.cpp
rectified95: I just used formatting that I find clearer, with more spacing between if-else clauses. It's not our repo, so I didn't know if we want to impose the same style

src/windows/SliderWindows/SliderWindows.vcxproj
rectified95: This is generated code. will adjust.

src/windows/SliderWindows/SliderView.cpp
rectified95: I didn't see examples of base64-encoded images in the examples here, so I just made sure both local and network files render properly. I don't think it's crucial that the thumb can be an svg

example/package.json
rectified95: copied from WebView, which we link to in our docs (https://microsoft.github.io/react-native-windows/docs/native-modules-setup#option-2-adding-windows-support-to-existing-sample-app)

example/metro.config.windows.js
rectified95: copied from WebView, which we link to in our docs (https://microsoft.github.io/react-native-windows/docs/native-modules-setup#option-2-adding-windows-support-to-existing-sample-app)

rectified95: Will delete

example/windows/SliderTestWindows/Package.appxmanifest
rectified95: will do

example/package.json
rectified95: changed.

example/windows/SliderTestWindows/App.cpp
rectified95: The same thing got generated for Tatiana: https://github.com/react-native-community/progress-view/pull/37/files#diff-c0553c5c1db81914e0098a7cf30ac5c3

src/windows/SliderWindows/SliderView.cpp
rectified95: The same formatting was used in the ProgressView and Picker repos when adding Windows support.

example/windows/SliderTestWindows/App.cpp
rectified95: Can you point me to the updated template you're referencing?

asklar: @rectified95 https://github.com/microsoft/react-native-windows/blob/master/vnext/local-cli/templates/cpp/src/App.cpp
https://github.com/microsoft/react-native-windows/blob/master/vnext/local-cli/templates/cpp/src/MainPage.cpp

README.md
asklar: * only sad reacts *

src/windows/SliderWindows/SliderView.cpp
asklar: i don't think it's crucial - but maybe clarify this in the documentation that you can't use base64 SVG as an image for this

example/windows/SliderTestWindows/App.cpp
rectified95: Thanks for the pointer,
I've looked at those files and they contain elements (eg. xamlNamespaceCpp) that get replaced by the CLI during project creation. It'd be difficult for me to replace them all manually.

This is just an example app, so I think it should be ok? Especially given that this is what was generated for me on RNW 62.

src/windows/SliderWindows/SliderView.cpp
rectified95: Updated the doc to say svg won't work.

README.md
```diff
@@ -308,17 +319,60 @@ To fix formatting errors, run the following:
...
+- (on iOS) Run `npx pod-install` from the `example` directory
+- (on Windows) You need to manually link the Slider module to your project.
+
+## Start the app
+
+- Run `yarn run:android` to run on Android
+- Run `yarn run:ios` to run on iOS
+- Run `yarn run:web` to run on web
+
+### Windows
```
asklar, line 340:
```suggestion
```

remove to make Windows just another one of the gang :)

rectified95:
@wojteg1337 @michalchudziak 
My colleague has signed off on this one. Can you guys take a look as well?

rectified95:
@michalchudziak Thanks for the approval!
Before you merge it, I just wanted call out the fact that **I bumped up the version of RN to 62**, and took a guess as to this being a breaking change, which would result in the release number getting change to something like 3.1.X (indicated in README). 
Let me know if this is fine, or whether you'd prefer to keep this module on an older version of React Native, with a note that on Windows the latest is required.

rectified95:
@michalchudziak What do you think about my last comment? :)

rectified95:
@michalchudziak @wojteg1337 Can you merge this one, plz? Do you want to address the question from my previous comments?

PR #221, opened 2020-08-26
fixes(web): sliding on track edge bugs

### LINKED ISSUE #199: [BUG][WEB] wrong value when reaching track edges
## Description
The web version has problems when during a drag the cursor goes over the left or right edge of the track.

Screen Recording 2020-07-03 at 13 43 14

## Reproducible Demo

See this [snack](https://snack.expo.io/@misterdev/[bug]-react-native-slider-drag-to-the-edge)

NOTE: the snack starts with an error message that I'm not able to fix, making a simple edit to the code updates the view and the error disappears
Screenshot 2020-07-03 at 14 47 52

misterdev:
Summary:
---------
Fixes #199 

When sliding, the delta is calculated using `e.locationX` ([code](https://github.com/react-native-community/react-native-slider/blob/master/src/js/RNCSliderNativeComponent.web.js#L278)) that is not aware of which is the element with the `onResponderMove` listener (see [here](https://github.com/necolas/react-native-web/issues/693#issuecomment-365125740)).

This causes #199 because when you slide over the edge of the track `locationX` contains the position of the cursor relative to an element outside of the slider, resulting in an incorrect calculation of the slider value.

In order to make it work, I used `e.pageX` to get the position of the cursor on the page and I compare it with the `x` of the slider container (`containerPositionX`).

`containerPositionX` is calculated when there's a slider layout change (using `onLayout`) and when there's a window resize

This works with `inverted` but not with `vertical` (that doesn't seem to be an implemented feature)

Test Plan:
----------

I don't know how to programmatically test it, I've manually tested it on the `examples`.

|Before|After|
|-|-|
|ezgif com-video-to-gif (1)|ezgif com-video-to-gif|

### DIFF
--- a/src/js/RNCSliderNativeComponent.web.js
+++ b/src/js/RNCSliderNativeComponent.web.js
@@ -6,6 +6,7 @@
  *
  */
 
+import ReactDOM from 'react-dom';
 import React, {useCallback} from 'react';
 import {View, StyleSheet} from 'react-native';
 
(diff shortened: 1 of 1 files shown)


alfonsocj:
@michalchudziak Do you mind checking this one? It's a bug that confuses web users a lot.

michalchudziak (approved):

PR #222, opened 2020-08-26
fixes(web): sliding over track edges bugs

misterdev:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "private": true,
   "workspaces": ["src", "example"],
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,7 @@
 {
   "private": true,
   "workspaces": ["src", "example"],
+  "version": "1.0.0",
```
misterdev, line 4: I think it's needed by yarn to install this as a package https://github.com/yarnpkg/yarn/issues/6195

PR #233, opened 2020-09-30
Add tapToSeek prop, fix unintentional scrolling, call onValueChanged when tapping

smallsaucepan:
Summary:
---------

This feature fork started out adding a tapToSeek prop to allow us to disable that functionality on iOS. The main motivation for this was to avoid inadvertent scrolling on touch down on iOS. Found what may be causing the inadvertent scrolling and fixed that (simulator only unfortunately - I don't a physical device). Also added the originally requested tapToSeek prop to make this functionality opt in on iOS.

Believe this fixes #156, fixes #211, fixes #212 and maybe fixes #147 as well.

Test Plan:
----------

Locally run yarn test passes.
CircleCI tests pass.
Visually inspect on iOS simulator and Android emulator. 
Tweaked one of the examples to demonstrate the tapToSeek behaviour.

PR #235, opened 2020-10-14
iOS: Fix problem where slider returned to original position.

MarkCSmith:
If the slide was quick, the new tap gesture recognizer would be
invoked and cause the slider to return to its original position
(ignoring the slide).

Summary:
---------

This fixes a problem introduced by #233 where the slider sometimes returns to its original position after a slide. To reproduce the problem, do a quick slide (short distance) with a slider that has `tapToSeek=true`.

Test Plan:
----------

Tested interactively on iOS and Android.

### DIFF
--- a/src/ios/RNCSliderManager.m
+++ b/src/ios/RNCSliderManager.m
@@ -11,9 +11,11 @@
 #import <React/RCTEventDispatcher.h>
 #import "RNCSlider.h"
 #import <React/UIView+React.h>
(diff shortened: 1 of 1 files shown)


agnieszkabugla:
@MarkCSmith , @michalchudziak 
The fix has been merged to master, but not released yet, right? I don't see property "tapToSeek" on the slider component in version 3.0.3. When is it planned to be released? :)

Issue #220, opened 2020-08-25
Lint error: Calling new methods (isAccessibilityFocused) on older versions


halaei:
## Environment

    "@react-native-community/slider": "^3.0.3",

```
System:
    OS: Linux 4.15 Ubuntu 18.04.5 LTS (Bionic Beaver)
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 203.29 MB / 15.55 GB
    Shell: 4.4.20 - /bin/bash
  Binaries:
    Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
    Yarn: 1.21.1 - ~/.yarn/bin/yarn
    npm: 6.14.5 - ~/.nvm/versions/node/v12.18.2/bin/npm
    Watchman: 20200712.162157.0 - /usr/local/bin/watchman
  SDKs:
    Android SDK:
      API Levels: 28, 29
      Build Tools: 28.0.3, 29.0.2, 29.0.3, 30.0.1
      System Images: android-19 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
      Android NDK: Not Found
  IDEs:
    Android Studio: Not Found
  Languages:
    Java: 1.8.0_265 - /usr/bin/javac
    Python: 2.7.17 - /home/hamid/.pyenv/shims/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: ^0.63.2 => 0.63.2 
  npmGlobalPackages:
    *react-native*: Not Found
```
## Description
Running `android/gradlew lint` reports this issue. I don't have enough information about whether it is a real bug or a false alarm needed to be suppressed by  @TargetApi(21):

../../src/main/java/com/reactnativecommunity/slider/ReactSlider.java:113: Call requires API level 21 (current min is 16): android.view.View#isAccessibilityFocused

```java
 110   public void onPopulateAccessibilityEvent(AccessibilityEvent event) {
 111     super.onPopulateAccessibilityEvent(event);
 112     if (event.getEventType() == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED ||
 113         (event.getEventType() == AccessibilityEvent.TYPE_VIEW_SELECTED && this.isAccessibilityFocused())) {
 114       this.setupAccessibility();
 115     }
 116   }
```
...
[...]

PR #248, opened 2020-11-17
Fix accessibility-related crash on Android 4.x (#220).

MarkCSmith:
Avoid calling isAccessibilityFocused() on Android < Lollipop.

Summary:
---------

Fixes issue #220.

Test Plan:
----------
We reproduced a crash on an Android 4.1 emulator, applied this patch, and observed that the crash no longer occurred.

### DIFF
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -109,9 +109,12 @@ public class ReactSlider extends AppCompatSeekBar {
   @Override
   public void onPopulateAccessibilityEvent(AccessibilityEvent event) {
     super.onPopulateAccessibilityEvent(event);
-    if (event.getEventType() == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED ||
-        (event.getEventType() == AccessibilityEvent.TYPE_VIEW_SELECTED && this.isAccessibilityFocused())) {
-      this.setupAccessibility();
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+      if (event.getEventType() == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED ||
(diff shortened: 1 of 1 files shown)


MarkCSmith:
> Hey @MarkCSmith ! Thank you for this PR.
> 
> I wonder if removing a call of `this.setupAccessibility();` at all on Android < Lollipop is the way to go. You claim that calling `isAccessibilityFocused()` is crashing an app. Perhaps we should omit just this call on these Android versions and just rely on `event.getEventType() == AccessibilityEvent.TYPE_VIEW_SELECTED`? Perhaps it'd resolve the crash and maintain `this.setupAccessibility();` call?

I think that approach would be fine.  However, my understanding is that React Native 0.64.x drops support for Android 4.x.  Maybe in that case it is no longer important to fix this issue.  What do you think?

src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -109,9 +109,12 @@ void setAccessibilityIncrements(List<String> accessibilityIncrements) {
...
   public void onPopulateAccessibilityEvent(AccessibilityEvent event) {
     super.onPopulateAccessibilityEvent(event);
-    if (event.getEventType() == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED ||
-        (event.getEventType() == AccessibilityEvent.TYPE_VIEW_SELECTED && this.isAccessibilityFocused())) {
-      this.setupAccessibility((int) mValue);
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+      if (event.getEventType() == AccessibilityEvent.TYPE_VIEW_ACCESSIBILITY_FOCUSED ||
+          (event.getEventType() == AccessibilityEvent.TYPE_VIEW_SELECTED && this.isAccessibilityFocused())) {
+          this.setupAccessibility((int)mValue);
```
BartoszKlonowski, line 116: I allowed myself to rebase your changes @MarkCSmith and I adjusted them to the latest of `setupAccessibility` method.

PR #257, opened 2021-01-15
Update project config to enable autolinking on Windows.

rectified95:
Summary:
---------

Autolinking has been implemented in React Native for Windows. Updating the module config to use it.

Additionally, I'm fixing the version number of RNW in package.json - currently it results in a 0.62rc being used, even though multiple newer releases have been made. It's necessary to use the later versions, since they contain a build fix for a problem with Folly and Visual Studio.

Test Plan:
----------
Created a new RNW app, added this module `yarn add @react-native-community/slider@4.0.0-rc.2`
Autolinking is applied when running `npx react-native run-windows`
When first building with Visual Studio, run npx react-native autolink-windows, then build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ or
 
 If using iOS, install cocoapods: `npx pod-install`
 
-On Windows, [`manually link the module`](#Manual-linking-of-the-module-on-Windows) to your app (autolinking will be released in RNW 63). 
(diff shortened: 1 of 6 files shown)


michalchudziak:
Thank you for this PR :)

michalchudziak (approved):

PR #260, opened 2021-01-26
Fix: Slider default step causes crash on Windows OS

### LINKED ISSUE #256: Slider manipulation results in crashes for react-native-windows when no step is specified
## Environment

* Building on a PC with visual studio 2019
* Using react-native-windows@0.64.0-preview.5
* Using @react-native-community/slider@4.0.0-rc.2

## Description

If you don't specify a step value, no default is set. This then causes xaml range value component to generate errors

## Reproducible Demo

* Setup a react-native-windows project
* Add @react-native-community/slider@4.0.0-rc.2 as a dependency and register the package in your project's app.cpp file
* Add a slider to your app, avoid specifying a step value
* Manipulate the slider

#### Observed behavior
The application crashes because no step value is set

#### Expected behavior
The application falls back to using a default step value

BartoszKlonowski:
This pull request fixes #256 

Summary:
---------

This PR provides the Slider with the fix for an application crashing when moving the Slider with a default value set for `<Step/>`.

The implementation introduces the similar approach to step's calculations as is already done for other platforms
(see implementation for [iOS](https://github.com/callstack/react-native-slider/blob/e7dd6eec8796832487afb886719ed622f575e5f6/src/ios/RNCSliderManager.m#L59-L61) and for [Android](https://github.com/callstack/react-native-slider/blob/e7dd6eec8796832487afb886719ed622f575e5f6/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java#L183-L185)).

The calculations done on Windows implementation returns the default step as 1% of available range: minimum to maximum.
NOTE: The implementation returns the calculated percentage value **in case of** original step being set to 0. This considers not only the default value, but also custom user value set to 0. This is intentional - during the analysis it turned out, that setting the value of slider's step to custom 0 the application also crashes with the same error.

For more information please see the commit messages.

Test Plan:
----------

Changes were tested manually with following scenarios:

**1.** Default step, minimumValue = 20, maximumValue = 140:
SliderDefault20-140

**2.** Default step, default minimum and maximum values:
SliderDefault0-1

**3.** Regression tests - step set at custom value of *1.234*, minimumValue = 34, maximumValue = 153
SliderCustom

### DIFF
--- a/src/windows/SliderWindows/SliderView.cpp
+++ b/src/windows/SliderWindows/SliderView.cpp
@@ -96,11 +96,9 @@ namespace winrt::SliderWindows::implementation {
                 if (propertyValue.IsNull()) {
                     this->ClearValue(xaml::Controls::Slider::StepFrequencyProperty());
                 }
-                else if (propertyValue.AsDouble() != 0) {
-                    this->StepFrequency(propertyValue.AsDouble());
-                }
                 else {
-                    this->StepFrequency(c_stepDefault);
+                    auto&& step = CalculateStepFrequencyPercentageValue(propertyValue.AsDouble());
+                    this->StepFrequency(step);
                 }
             }
             else if (propertyName == "inverted") {
@@ -235,6 +233,15 @@ namespace winrt::SliderWindows::implementation {
                     }
                     eventDataWriter.WriteObjectEnd();
                 });
+        }
+    }
+
+    const double SliderView::CalculateStepFrequencyPercentageValue(const double& stepPropertyValue) const noexcept {
+        if (stepPropertyValue != 0) {
+            return stepPropertyValue;
+        }
+        else {
+            return (m_maxValue - m_minValue) / 100.f;
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ Callback continuously called while the user is dragging the slider.
 
 Step value of the slider. The value should be between 0 and (maximumValue - minimumValue). Default value is 0.
 
+On Windows OS the default value is 1% of slider's range (from `minimumValue` to `maximumValue`).
+
 | Type   | Required |
 | ------ | -------- |
 | number | No       |
--- a/src/windows/SliderWindows/SliderView.h
+++ b/src/windows/SliderWindows/SliderView.h
@@ -37,7 +37,7 @@ namespace winrt::SliderWindows::implementation {
 
         const double CalculateStepFrequencyPercentageValue(const double& stepPropertyValue) const noexcept;
 
-        int64_t m_maxValue, m_minValue;
+        double m_maxValue, m_minValue;
         double m_value;
     };
 }
--- a/src/windows/SliderWindows/SliderView.h
+++ b/src/windows/SliderWindows/SliderView.h
@@ -34,9 +34,11 @@ namespace winrt::SliderWindows::implementation {
         void OnManipulationCompletedHandler(
             winrt::Windows::Foundation::IInspectable const& sender,
             xaml::Input::ManipulationCompletedRoutedEventArgs const& args);
-        
-        double m_value, m_maxValue, m_minValue;
-        const double c_stepDefault = 0.1;
+
+        const double CalculateStepFrequencyPercentageValue(const double& stepPropertyValue) const noexcept;
+
+        int64_t m_maxValue, m_minValue;
+        double m_value;
     };
 }
 


src/windows/SliderWindows/SliderView.cpp
```diff
@@ -97,7 +97,8 @@ namespace winrt::SliderWindows::implementation {
                     this->ClearValue(xaml::Controls::Slider::StepFrequencyProperty());
                 }
                 else {
-                    this->StepFrequency(propertyValue.AsDouble());
+                    auto&& step = CalculateStepFrequencyPercentageValue(propertyValue.AsDouble());
```
BartoszKlonowski, line 100: The advantage of that approach (comparing to just making a constant default step) is that it provides the `<Slider/>` with the same resolution no matter what length is the `<Slider/>`.

wojteg1337:
@BartoszKlonowski It looks like you have conflicts

wojteg1337:
Imo I like this solution more than [MR](https://github.com/callstack/react-native-slider/pull/259/files)

wojteg1337 (approved):

BartoszKlonowski:
@wojteg1337 Thanks!  I will solve the conflicts tomorrow, by rebasing my changes and I will keep the type conversion fix implemented in the merged solution,

src/windows/SliderWindows/SliderView.cpp
```diff
@@ -96,11 +96,9 @@ namespace winrt::SliderWindows::implementation {
                 if (propertyValue.IsNull()) {
                     this->ClearValue(xaml::Controls::Slider::StepFrequencyProperty());
                 }
-                else if (propertyValue.AsDouble() != 0) {
-                    this->StepFrequency(propertyValue.AsDouble());
-                }
                 else {
-                    this->StepFrequency(c_stepDefault);
+                    auto&& step = CalculateStepFrequencyPercentageValue(propertyValue.AsDouble());
```
rectified95, line 100: Does assigning to an r-value reference provides any benefit if you don't use std::move?

src/windows/SliderWindows/SliderView.cpp
```diff
@@ -237,4 +235,13 @@ namespace winrt::SliderWindows::implementation {
                 });
         }
     }
+
+    const double SliderView::CalculateStepFrequencyPercentageValue(const double& stepPropertyValue) const noexcept {
```
rectified95, line 239: there's no point in passing a primitive type by reference if you don't modify the incoming parameter

src/windows/SliderWindows/SliderView.cpp
```diff
@@ -237,4 +235,13 @@ namespace winrt::SliderWindows::implementation {
...
+        if (stepPropertyValue != 0) {
+            return stepPropertyValue;
+        }
+        else {
+            return (m_maxValue - m_minValue) / 100.f;
```
rectified95, line 244: don't need '.f' since m_maxvalue is already a floating-point type

src/windows/SliderWindows/SliderView.cpp
rectified95: While marking the function as an accessor with the latter 'const', I don't think that making the return type 'const double' changes anything

src/windows/SliderWindows/SliderView.cpp
rectified95: Good idea! Worth noting, though, that this assumes min/max values are specified, which they aren't always.

rectified95:
It's probably a good idea to parametrize the default step value! However, since it relies on defaults for min/maxValue, you could consider providing defaults for those 2 props as well. Otherwise, for cases when they're not specified by the user, you end up returning 0, which crashes the app, and is what you're trying to fix here in the first place.

BartoszKlonowski:
@rectified95 Thanks for the review!
You've got the point about the default min & max values - this should be covered.
I'll file an issue about it, so it will be handled ASAP.

That issue I'll file can also be the place to cover some of your comments.

---

As for the:

> I don't have the privileges to assign issues in this repo, so couldn't self-assign, but still wanted to fix the initial implementation I contributed here :)

You can always ask the owners to assign you to the item, or at least write a comment saying "I'll work on this". There will be no problems assigning you to it 👍🏻

rectified95:
Cool, so I double checked it and it doesn't crash. I remember there being issues with using the static `.defaultProps` on RN components in the past - must have been fixed since then. 

And I agree that if someone misused the component without it crashing that's probably alright.

src/windows/SliderWindows/SliderView.cpp
BartoszKlonowski: `auto&&` will bind to both r-values and l-values. It is not always and only r-value reference. Scott Meyers in his book used to call it "*universal reference*".
Let me know if I understand your comment correctly.

src/windows/SliderWindows/SliderView.cpp
BartoszKlonowski: Well... I've implemented it when there was still no your parallel fix delivered, so both `m_maxValue` and `m_minValue` were integers, which resulted in warning without `.f` modifier.
You've got the point here - this is a merge mistake done by me, because when I was merging our solutions I completely forgot that you have changed the types of min and max values (though I intentionally left that part of your delivery as very usefull change).

src/windows/SliderWindows/SliderView.cpp
rectified95: I was imprecise, and std::move is relevant here. My point was that while `auto&&` can bind both l- and r-value refs, what the CalculateStep function returns is an r-value. As such, it'd be enough to just write 
`auto step = ..` since you're just capturing an ephemeral value. So in this case, following Meyer's explanation, `auto&& step` gets resolved to an l-value reference, but I'm not sure this is needed, since this new reference doesn't really refer to anything after the function returns the rvalue.

tl;dr; I just meant you can skip `&&`

Issue #271, opened 2021-03-15
minimumTrackImage, maximumTrackImage, and trackImage do not work together 


YeetOrBeYate:
## Environment

System:
    OS: macOS 10.15.7
    CPU: (4) x64 Intel(R) Core(TM) i5-7400 CPU @ 3.00GHz
    Memory: 1008.80 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.15.3 - /var/folders/1t/_jt97n35417ff5hsqd0m7dbw0000gn/T/yarn--1615841103810-0.40424128651417157/node
    Yarn: 1.22.4 - /var/folders/1t/_jt97n35417ff5hsqd0m7dbw0000gn/T/yarn--1615841103810-0.40424128651417157/yarn
    npm: 6.14.9 - ~/.nvm/versions/node/v14.15.3/bin/npm
    Watchman: Not Found
  Managers:
    CocoaPods: Not Found
  SDKs:
    iOS SDK:
      Platforms: iOS 14.4, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
    Android SDK: Not Found
  IDEs:
    Android Studio: 4.0 AI-193.6911.18.40.6626763
    Xcode: 12.4/12D4e - /usr/bin/xcodebuild
  Languages:
    Java: Not Found
    Python: 3.8.5 - /Users/kylecorbin/.pyenv/shims/python
  npmPackages:
    @react-native-community/cli: Not Found
    expo: ^40.0.0
    react: 16.13.1 => 16.13.1 
    react-native: https://github.com/expo/react-native/archive/sdk-40.0.0.tar.gz => 0.63.2 
    react-native-macos: Not Found
    expo-cli: 4.0.16
  npmGlobalPackages:
    *react-native*: Not Found
✨  Done in 26.91s.

## Description

This looks like a new issue, please direct me to another thread if there is one.

Only trying this on the ios version. Setting any of the above minimumTrackImage/maximumTrackImage props  _along with_ a trackImage prop does not work. Doesn't matter if its just one minimumTrackImage/maximumTrackImage or both, they don't work. Even when I try to set a minimumTrackImage/maximumTrackImage props _without_ setting a trackImage prop, neither works. 

Is minimumTrackImage/maximumTrackImage meant to be used in conjunction of a trackImage? Are they meant to work without one? Is using all three part of the problem?

## Reproducible Demo

```
import React from 'react';
[...]

PR #275, opened 2021-04-06
Fix minimumTrackImage, maximumTrackImage, and trackImage do not work together

dstdnk:
Summary:
---------

Fixing issue [#271 ](https://github.com/callstack/react-native-slider/issues/271) **only** for iOS.

Behavior after fix:
- `trackImage`
Screen Shot 2021-04-06 at 8 14 11 PM

- `trackImage` and `minimumTrackImage` 
Screen Shot 2021-04-06 at 8 17 10 PM

- `trackImage` and `maximumTrackImage`
Screen Shot 2021-04-06 at 8 17 23 PM

- `trackImage and `minimumTrackImage` and `maximumTrackImage`
Screen Shot 2021-04-06 at 8 17 36 PM

Test Plan:
----------

Manually tested with other properties - works as before the changes.

### DIFF
--- a/src/ios/RNCSlider.h
+++ b/src/ios/RNCSlider.h
@@ -23,8 +23,6 @@
 @property (nonatomic, strong) UIImage *maximumTrackImage;
 @property (nonatomic, strong) UIImage *thumbImage;
 @property (nonatomic, assign) bool tapToSeek;
-@property (nonatomic, assign) bool minimumTrackImageSet;
-@property (nonatomic, assign) bool maximumTrackImageSet;
 @property (nonatomic, strong) NSString *accessibilityUnits;
 @property (nonatomic, strong) NSArray *accessibilityIncrements;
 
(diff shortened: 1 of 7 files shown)


src/ios/RNCSlider.h
```diff
@@ -23,6 +23,8 @@
 @property (nonatomic, strong) UIImage *maximumTrackImage;
 @property (nonatomic, strong) UIImage *thumbImage;
 @property (nonatomic, assign) bool tapToSeek;
+@property (nonatomic, assign) bool minimumTrackImageSet;
+@property (nonatomic, assign) bool maximumTrackImageSet;
```
michalchudziak, line 27: I don't think we should make these public - let's keep them private.

michalchudziak (changes requested):

src/ios/RNCSlider.h
dstdnk: Done.

src/ios/RNCSlider.m
```diff
@@ -10,7 +10,8 @@
 @implementation RNCSlider
 {
   float _unclippedValue;
-}
+  bool _minimumTrackImageSet;
+  bool _maximumTrackImageSet;}
```
michalchudziak, line 14: nit: let's move curly bracer to the next line

michalchudziak (approved):

Issue #264, opened 2021-02-12
Impossible to change slider position with VoiceOver if it has less than 6 steps configured.


elfstyle:
## Environment

react: 16.13.1 
react-native: 0.63.4

## Description

If the slider is configured with the `step` prop and a number of steps in [min, max] range is less than 6 then it is impossible to change the value using VoiceOver.

## Reproducible Demo
// numberOfSteps = (maximumValue-minimumValue)/step

// Can change the value if  numberOfSteps>=6
`<Slider minimumValue={0} maximumValue={6} step={1} />`

// **Can only increase** the value if  numberOfSteps===5
`<Slider minimumValue={0} maximumValue={5} step={1} />`

// **Can't move slider** if  numberOfSteps<5
`<Slider minimumValue={0} maximumValue={4} step={1} />`

dstdnk:
Hi @elfstyle, are you using bare `react-native` setup, or `expo`?

elfstyle:
> Hi @elfstyle, are you using bare `react-native` setup, or `expo`?
> Are you able to reproduce the issue with the [example app](https://github.com/callstack/react-native-slider/tree/master/example)?

Bare "react-native": "0.63.4"

elfstyle:
Can reproduce on freshly created app https://github.com/elfstyle/SliderA11Y

dstdnk:
@elfstyle unfortunately I could not reproduce your issue.
I used:
```
"@react-native-community/slider": "3.0.3",
"react": "16.13.1",
"react-native": "0.63.4"
```

Here's what I did:
```
npx react-native init reproduce_264 --version 0.63.4 
cd reproduce_264
yarn add @react-native-community/slider
npx pod-install
xed -b ios
```
In the `App.js` I've added:
```
<View
  style={{
    padding: 20,
    height: 300,
    justifyContent: 'space-between',
  }}>
  <Slider minimumValue={0} maximumValue={6} step={1} />   // should work
  <Slider minimumValue={0} maximumValue={5} step={1} />   // should only increase
  <Slider minimumValue={0} maximumValue={4} step={1} />   // should not work
</View>
```
I turned on VoiceOver, tap the slider once to activate it, then double-tap and hold.
While holding I move the slider.

Please let me know if you're able to reproduce the problem with the steps above.

dstdnk:
> Here is the recording
[...]

PR #281, opened 2021-04-30
Fix bug, caused by fixing issue #264.

dstdnk:
Summary:
After fixing the issue #264, the new bug pops up and was left unnoticed.
If the slider is continuously scrolled, the slider thumb jumps sporadically.

https://user-images.githubusercontent.com/12425395/116679984-98f7e880-a9c4-11eb-95ab-e80240a20c70.mov

Now the bug is fixed.

Test Plan:
Manually test that the slider works fine while sliding, and with VoiceOver.


michalchudziak (approved):
nit: remove whitespaces

PR #308, opened 2021-08-29
Fix: Remove redundant React, Rreact Native & React Native Windows from `devDependencies`

### LINKED ISSUE #307: Slider 4.1.3 requires React 16.11.0 in conflict with 17.0.2
Hello world :-)

I have upgraded React to 17.0.2 and React Native to 0.65.1. Now I cannot install `@react-native-community/slider` 4.1.3 with npm as it requires React 16.11.0..

Please update :-)

BartoszKlonowski:
This pull request fixes #307 

It removes the React, React Native and React Native Windows from `devDependencies` when it comes to *src/package.json* - package dependencies.

The reason that these can be safely removed is that these libraries are also listed in the `peerDependencies` which makes them still being checked when installing.
Both workspaces should not be affected as *src/* should use the packages provided by the consumer, while *example/* should satisfy `peerDependencies` by having them installed in the example app
(moreover: *example* is about to change in upcoming deliveries).

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -36,9 +36,6 @@
     "flow-bin": "0.113.0",
     "jest": "^24.9.0",
     "metro-react-native-babel-preset": "^0.56.0",
-    "react": "^16.11.0",
-    "react-native": "^0.62.3",
-    "react-native-windows": "^0.62.0-0",
     "react-test-renderer": "16.9.0"
   },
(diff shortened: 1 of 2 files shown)


deorst (approved):

cederom:
TANK U =)

PR #323, opened 2021-10-04
ISSUE-319-Accessibility not working perfectly in Android 11 devices 

srinu6:
Summary:
---------

- For Android 11.0 devices, while sliding we are only hearing the percentage of the slider but not `accessibilityIncrements` and `accessibilityUnits`. It is working fine for the other versions below Android 11.
- As part of this PR we are fixing this issue in Android 11 devices.
- I have tested this in various android versions(Android 9, Android 10, Android 11) devices with this fix, it is working perfectly now.
### Before - Android version 11:
https://drive.google.com/file/d/1mfqT6bfyoVfSETyN89gtEo3golnJKiId/view?usp=sharing

### After  - Android version 11:
https://drive.google.com/file/d/1ADx0VLZxxtxni-_MozDAyNBLu2cS_HwJ/view?usp=sharing

### After - Android version 10:
https://drive.google.com/file/d/1nL1e9410y6pdMmjFUXuImYJWqz8Xqrbl/view?usp=sharing

Test Plan:
----------

- Changes were tested manually in different Android versions of devices.

### DIFF
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
@@ -152,7 +152,7 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
     if (view.isAccessibilityFocused() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
       int index = (int)value;
       view.setupAccessibility(index);
-    }  
+    }
   }
 
   @ReactProp(name = "minimumValue", defaultDouble = 0d)
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
@@ -151,13 +151,7 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
     view.setOnSeekBarChangeListener(ON_CHANGE_LISTENER);
     if (view.isAccessibilityFocused() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
       int index = (int)value;
-      String sliderValue = view.mAccessibilityIncrements.get(index);
-      int stringLength = view.mAccessibilityUnits.length();
-      String spokenUnits = view.mAccessibilityUnits;
-      if (sliderValue != null && Integer.parseInt(sliderValue) == 1) {
-        spokenUnits = spokenUnits.substring(0, stringLength - 1);
-      }
-      view.announceForAccessibility(String.format("%s %s", sliderValue, spokenUnits));
+      view.setupAccessibility(index);
     }  
   }
 
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
@@ -149,18 +149,16 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
     view.setOnSeekBarChangeListener(null);
     view.setValue(value);
     view.setOnSeekBarChangeListener(ON_CHANGE_LISTENER);
-    if (view.isAccessibilityFocused()) {
-      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-        int index = (int)value;
-        String sliderValue = view.mAccessibilityIncrements.get(index);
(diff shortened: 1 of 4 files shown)


srinu6:
If we change `defaultCompileSdkVersion` to API 30 then this CircleCI build issue will be fixed.

srinu6:
> Hello @srinu6, thank you very much for this pull request! I've linked the issue which is to be fixed by your changes. Let me come back to this as soon as I'll find time to review.
> 
> In the meantime there are two things about your pull request:
> 
> * It's always a good practice to use a GIFs or any kind of demo that wouldn't require entering a link.
>   Of course it's a great idea to provide a demo in the first place, thank you!
> * The build of Android fails due to line 152 of your changes using `Build.VERSION_CODES`.
> 
> > If we change `defaultCompileSdkVersion` to API 30 then this CircleCI build issue will be fixed.
> 
> Yes, exactly and I was in the middle of sending a PR with that while writing my response to your PR. I'll let you know once it's delivered, so you can rebase and check your changes with updated SDK.

Thank you @BartoszKlonowski

BartoszKlonowski:
@srinu6 The `compileSdkVersion` is updated - feel free to rebase and check.

srinu6:
> @srinu6 The `compileSdkVersion` is updated - feel free to rebase and check.

Thank you @BartoszKlonowski, just rebased, passed all the checks. Can you please verify this?

src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -149,6 +149,16 @@ public void setValue(ReactSlider view, double value) {
     view.setOnSeekBarChangeListener(null);
     view.setValue(value);
     view.setOnSeekBarChangeListener(ON_CHANGE_LISTENER);
+    if (view.isAccessibilityFocused() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
```
BartoszKlonowski, line 152: DRY. The implementation below is identical to what is already implemented here:
https://github.com/callstack/react-native-slider/blob/269e421bc89137cdb774d0e02c9133e6f208de21/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java#L142
Perhaps we could use that so that the `ReactSlider` will handle all that with his private members?
As an example please take a look at a suggested change below.

src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -149,6 +149,16 @@ public void setValue(ReactSlider view, double value) {
...
+      int index = (int)value;
+      String sliderValue = view.mAccessibilityIncrements.get(index);
+      int stringLength = view.mAccessibilityUnits.length();
+      String spokenUnits = view.mAccessibilityUnits;
+      if (sliderValue != null && Integer.parseInt(sliderValue) == 1) {
+        spokenUnits = spokenUnits.substring(0, stringLength - 1);
+      }
+      view.announceForAccessibility(String.format("%s %s", sliderValue, spokenUnits));
```
BartoszKlonowski, line 160: As a follow up to comment above this is how it would look like if utilized the `setupAccessibility`.
```suggestion
      int index = (int)value;
      view.setupAccessibility(index);
```

Of course, the `setupAccessibility` should be refactored so that it takes an argument of index to use, and the call of this method should be then modified to use `(int)mValue` when called in `ReactSlider` class or `(int)value` when called here.

BartoszKlonowski (changes requested):
The idea is great and it works 👍, but to make this PR deliverable the DRY should be fixed.
Please take a look at my suggestion to use the refactored `setupAccessibilit` of `ReactSlider` class.

Feel free to reach out to me if you have any questions or if you disagree.

src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
srinu6: modified as per suggestion, thank you!

src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
srinu6: modified, thank you!

BartoszKlonowski (approved):
Thank you, @srinu6!

srinu6:
Hi @BartoszKlonowski , when the latest version will be released?

BartoszKlonowski:
Today, within one hour.

PR #349, opened 2022-01-17
`value` property implemented as a controlled value

### LINKED ISSUE #192: How to update value visually by changing prop?
## Ask your Question

We want to update the slider visually while a track is playing, we want to be able to read from a constantly updated position. This seemed to work with the react native slider but not with this one, any ideas?

Thanks and awesome work!

Our Slider code looks like this:
```
<Slider
        style={{ marginRight: 10, marginLeft: 10, flex: 1 }}
        minimumValue={0}
        maximumValue={1}
        value={
          this.props.seekSliderPosition ? this.props.seekSliderPosition : 0
        }
        onValueChange={value => this.props.onSeekSliderValueChange(value)}
        onSlidingComplete={value =>
          this.props.onSeekSliderSlidingComplete(value)
        }
        disabled={
          this.props.playbackState === PLAYBACK_STATES.LOADING ||
          this.props.playbackState === PLAYBACK_STATES.ENDED ||
          this.props.playbackState === PLAYBACK_STATES.ERROR
        }
        minimumTrackTintColor={Colors.brandBlue}
        maximumTrackTintColor={Colors.backgroundTrackColor}
        thumbTintColor={Colors.brandBlue}
      />
```
On our store we have:
```
@computed
  get seekSliderPosition() {
    if (
      this.playbackInstance !== null &&
      this.playbackInstancePosition != null &&
      this.playbackInstanceDuration != null &&
      this.playbackInstanceDuration !== 0
    ) {
      return this.playbackInstancePosition / this.playbackInstanceDuration;
    }
    return 0;
  }
```

### LINKED ISSUE #292: Controlled value
Instead of having 'value' prop as initial value and then we cannot control it , I suggest to rename it to  'initialValue' to keep the same functionality ,  and name 'value' as 'controlledValue'. if 'controlledValue' isn't set it should work as it is now.

Example that this is useful : 
After onSlidingComplete I ask user if he is sure that he want to set the value at x % . If he press cancel , I can't control the value to set it back at previous state.

Solutions:
- Just add controlledValue prop without changing 'value' prop to be compatible with previous versions
- Refactor 'value' ->  'initialValue' and 'controlledValue' -> 'value'

### LINKED ISSUE #295: Slider jumps around when used to control react-native-video player.
Hi everyone,

I am using react-native-slider as a custom video seeker for react-native-video. 

I am updating the value prop to update the slider position according to the video progress.

But when I try to seek the video, using react-native-slider, and the video progress is updated as the video is running, the slider keeps jumping between the thumb position and the updated video progress.

I tried to handle this by not updating the slider value by setting  a flag on onSlidingStart event, but it is often delayed and the video progress updates before the flag is set.

I tried messing around with the native android code to handle this, but due to my limited knowledge of Android, I couldn't make much progress. Can you point me towars what changes I need to make to the native code for Android, so that the slider does not jump to the position of value prop while the sliding event is active.

Thanks and regards.

BartoszKlonowski:
**This pull request closes #292**

It introduces the improved control over the `value` property of the Slider and provides the user with the ability to control the thumb either by updating the `value` externally or by dragging/tapping the thumb.

(*This pull request also fixes #295 and closes #192*)

**NOTE:** This pull request affects all three platforms: iOS, Windows, Android

---

**How to use the new controlled `value` property?**

The `value` property should be considered **write-only**.
* It still does not need to be updated during dragging.
* Value written to that property will affect thumb's position immediately
* Updates to the `value` property **does not trigger** the `OnValueChange` event
What is programmatically written into `value` property should be tracked in the developer's implementation.
* Updates to the thumb's position (either by tap or drag) **will trigger** the `OnValueChange` event just like before

**What about the backward compatibility?**
It should not affect your current implementation.<br/>Unless you already had any update to the `value` prop implemented, you won't observe any change in the behavior of `<Slider/>`.
If you did have the `value` being updated somehow in your component, please adjust the implementation to separate the value of Slider and the value of your component.

In case of any problems or questions please report an issue.

---

**How does it work?**

To present the new controlled value let's use the example given in [this comment](https://github.com/callstack/react-native-slider/issues/295#issuecomment-893464169),
**but**
let's remove the `seeking` ref, we don't need it anymore :)

```javascript
const VideoComponent = () => {
  const [value, setValue] = useState(0);

  useEffect(() => {
    setInterval(() => {
      setValue((v) => v + 1);
    }, 50);
  }, []);

  const handleSeek = (val) => {
    setValue(val);
  };

  return (
    <View style={styles.container}>
      <Text style={styles.text}>{value}</Text>
      <Slider
        value={value}
        minimumValue={0}
        maximumValue={1000}
        step={1}
        minimumTrackTintColor="green"
        maximumTrackTintColor="red"
        onValueChange={(progress) => {
          console.log("value change", progress);
        }}
        onSlidingStart={(progress) => {
          console.log("sliding start at: %d", progress);
        }}
        onSlidingComplete={(progress) => {
          handleSeek(progress);
          console.log("sliding complete", progress);
        }}
      />
    </View>
  );
};
```
This will result in the following behavior:
SliderReproDrag

---

**Implementation details**

The `value` property was previously used as an initial value of the `<Slider/>`. Changing this property would however update the position of the thumb, but the main issue was that such update was in conflict with the "manual" dragging of the thumb.

So the Slider **was a controlled component** under the hood, but this feature was unreliable and was not working.

This implementation fixes several issues with the `value` property itself, but the main feature added is that `<Slider/>` will now differentiate between `value` updated externally (programmatically) and `value` updated by user's action.
**Priority will be for user's action.** Each time user manipulates the `<Slider/>`, **no** external value will be accepted.
For iOS the implementation also changed the `value` to be the custom field with sliding condition.

---

For more questions or in case of any problem please report an issue.

### DIFF
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -56,6 +56,8 @@ public class ReactSlider extends AppCompatSeekBar {
    */
   private double mValue = 0;
 
+  private boolean isSliding = false;
+
   /** If zero it's determined automatically. */
   private double mStep = 0;
 
@@ -97,6 +99,14 @@ public class ReactSlider extends AppCompatSeekBar {
   /* package */ void setStep(double step) {
     mStep = step;
     updateAll();
+  }
+
+  boolean isSliding() {
+    return isSliding;
+  }
+
+  void isSliding(boolean isSliding) {
+    this.isSliding = isSliding;
   }
 
   void setAccessibilityUnits(String accessibilityUnits) {
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
@@ -35,8 +35,6 @@ import javax.annotation.Nullable;
 
 /**
  * Manages instances of {@code ReactSlider}.
- *
- * Note that the slider is _not_ a controlled component.
  */
 public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
 
@@ -87,8 +85,7 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
           reactContext.getNativeModule(UIManagerModule.class).getEventDispatcher().dispatchEvent(
               new ReactSliderEvent(
                   seekbar.getId(),
-                  ((ReactSlider) seekbar).toRealProgress(progress),
-                  fromUser));
+                  ((ReactSlider)seekbar).toRealProgress(progress), fromUser));
         }
 
         @Override
@@ -98,7 +95,7 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
           reactContext.getNativeModule(UIManagerModule.class).getEventDispatcher().dispatchEvent(
               new ReactSlidingStartEvent(
                   seekbar.getId(),
-                  ((ReactSlider) seekbar).toRealProgress(seekbar.getProgress())));
+                  ((ReactSlider)seekbar).toRealProgress(seekbar.getProgress())));
         }
 
         @Override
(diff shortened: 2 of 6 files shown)


deorst (approved):

mosoakinyemi (approved):
Awesome work!

PR #353, opened 2022-01-31
Ignore all node_modules

BartoszKlonowski:
This pull request ignores all libraries in *node_modules* directory when it comes to flow checks.

Reasoning behind it:
* Some modules have their files malformed in purpose which leads to flow check fails for them, see facebook/flow#2364
* This fixes errors in dependabot PR #348 
* This speeds up the flow check of the sources in this library, reducing the CI minutes.

All comments against this change are welcomed.


mosoakinyemi (approved):
Looks good!

PR #364, opened 2022-03-28
fix: flow module resolution

deorst:
Summary:
---------

There was a `flow` related error - static type checker could not resolve dependency packages from `node_modules`.

Test Plan:
----------

No library functionality has changed.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -18,6 +18,7 @@
 .*/node_modules/@react-native-community/cli/.*/.*
 
 [include]
+../../node_modules
 
 [libs]
 node_modules/react-native/interface.js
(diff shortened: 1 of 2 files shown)


.flowconfig
```diff
@@ -18,6 +18,7 @@
 .*/node_modules/@react-native-community/cli/.*/.*
 
 [include]
+../../node_modules
```
BartoszKlonowski, line 21: Wouldn't it collide with line 9?
https://github.com/callstack/react-native-slider/blob/470288ca40b77cd538ed1c8101fe63f332d47d02/.flowconfig#L9

BartoszKlonowski (commented):
In the description you mention the flow related error, but what was this error?
Maybe the test plan could include the way to reproduce the error locally (I guess by launching flowcheck some certain way)?

thymikee:
This repo moved to TS, so the issue is not actionable anymore :)

PR #367, opened 2022-04-10
Fix: onChange is not called for tap when on Android

### LINKED ISSUE #366: Tap on slider is not triggering onValueChange
## Description

**Library Version:** 4.2.1

When you tap on slider value is not changed. 

`onValueChange`  is not triggered at all but UI changes accordingly

BartoszKlonowski:
This pull request fixes #366 

The solution is to:
* indicate the user action as a source of event depending on the current sliding status
Having the `fromUser` set to fixed value of `false` was causing the event to be recognized as avoidable.
* always send the onChange event no matter if `fromUser` is set
This check is now removed as it was found to be redundant and could introduce conflicts for the sliding vs taping.

Please see below presentation of the working fix:
OnChangeAndroidFixResult


BartoszKlonowski:
CC: @Hellhiem

Hellhiem (approved):
🚀

pdaus1405:
@BartoszKlonowski @Hellhiem 
when will a new version be released with this pr?

BartoszKlonowski:
@pdaus1405 I still have some of the fixes waiting to be merged, but the *v4.2.2* will be published not later than the end of this week (Friday).

PR #368, opened 2022-04-12
Fix: onAccessibilityAction causes crash

### LINKED ISSUE #250: iOS onAccessibilityAction bug
## Environment

info Fetching system and libraries information...
System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
    Memory: 21.64 GB / 31.72 GB
  Binaries:
    Node: 12.10.0 - ...\AppData\Local\Temp\yarn--1606148773529-0.05497837007190132\node.CMD
    Yarn: 1.22.0 - ////AppData\Local\Temp\yarn--1606148773529-0.05497837007190132\yarn.CMD
    npm: 6.10.3 - ....nodejs\npm.CMD
    Watchman: 4.9.4 - C:\bin\watchman.EXE
  SDKs:
    Android SDK:
      API Levels: 23, 28, 29
      Build Tools: 28.0.3, 29.0.3
      System Images: android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: Version  3.5.0.0 AI-191.8026.42.35.6010548
  npmPackages:
    react: 16.9.0 => 16.9.0
    react-native: ^0.61.5 => 0.61.5

## Description

When passing onAccessibilityAction to the slider, I see the following error: 
Exception thrown while executing the UI block: 1 [RNCSlider setOnAccessibilityActon:]:
unrecognized selector set to instance 0x7f9179fa4900.  

I'm passing onAccessibilityAction to this component because I'm unable to decrement the slider in accessibility mode in iOS.  

## Reproducible Demo

Here's the change hander: 

const handleAccessibleChangeMaxWaitTime = increment => {
    const times = [5, 15, 30, 45, null];
    let index = times.indexOf(maxWaitTime);
    if (index === -1) index = 4;
    if (increment) {
      if (index === times.length - 1 || !maxWaitTime) return;
      setLocationMaxWaitTime(times[index + 1]);
    } else {
      if (index === 0) return;
      setLocationMaxWaitTime(times[index - 1]);
    }
  };

Here's the onAccessibilityAction props I'm passing:

              accessibilityActions={[
                { name: 'increment' },
                { name: 'decrement' },
              ]}
              onAccessibilityAction={event => {
                switch (event.nativeEvent.actionName) {
                  case 'increment':
                    handleAccessibleChangeMaxWaitTime(true);
                    break;
                  case 'decrement':
                    handleAccessibleChangeMaxWaitTime(false);
                    break;
                  default:
                    break;
                }
              }}

I've observed this problem both on the current release and on the v4.0.0-rc.2.  Works fine on Android.

BartoszKlonowski:
This pull request fixes #250
It adds the explicit implementation of utilizing the `onAccessibilityAction` property passed to the Slider exposed to JS side.

The implementation **does not** overrides the implementation of the event itself, but it takes that property directly from props passed to Slider as part of `ViewProps`.
The extracted property is explicity passed to Slider component implementation.

Tested on simulator using the Accessibility Inspector from Xcode tools.

### DIFF
--- a/src/js/Slider.js
+++ b/src/js/Slider.js
@@ -11,7 +11,7 @@
 'use strict';
 
 import React from 'react';
-import {Image, Platform, StyleSheet} from 'react-native';
+import {Image, Platform, StyleSheet, AccessibilityActionEvent} from 'react-native';
 import RCTSliderNativeComponent from './RNCSliderNativeComponent';
 
 import type {Ref} from 'react';
@@ -286,7 +286,7 @@ const SliderComponent = (
       }
     : null;
   const onAccessibilityActionEvent = onAccessibilityAction
-    ? (event: Event) => {
+    ? (event: AccessibilityActionEvent) => {
         onAccessibilityAction(event.nativeEvent.value);
       }
     : null;
--- a/src/js/Slider.js
+++ b/src/js/Slider.js
@@ -11,7 +11,12 @@
 'use strict';
 
 import React from 'react';
-import {Image, Platform, StyleSheet, AccessibilityActionEvent} from 'react-native';
+import {
+  Image,
+  Platform,
+  StyleSheet,
+  AccessibilityActionEvent,
+} from 'react-native';
 import RCTSliderNativeComponent from './RNCSliderNativeComponent';
 
 import type {Ref} from 'react';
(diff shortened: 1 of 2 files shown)


BartoszKlonowski:
CC: @alexkramerbanjo

Issue #370, opened 2022-04-17
SeekBar on Android is rendering incorrectly in RTL mode


pavel-stryber:
## Environment

**Library Version:** 4.2.1

## Description

1. Slider renders incorrectly on Android in rtl mode. Images how it is now and should be are attached.
2. Currently slider changes it's layout direction based on device language. I want that it was relying on isRTL value of RN instead. It's important in case if we don't have translations for some rtl language and we have to use English as a default.

## Reproducible Demo

<table>
<tr>
<td>
Current behavior:
<img src="https://user-images.githubusercontent.com/44573208/163732628-060f6bd1-f40b-48f0-aec6-60ac1dc7ae22.png" />
</td>
<td>
Correct behavior
<img src="https://user-images.githubusercontent.com/44573208/163732631-dc3bce36-d3b4-481b-93c8-88e6684be01e.png" />
</td>
</tr>
</table>

pavel-stryber:
@BartoszKlonowski thank you for review. When are you going to make a new release?

BartoszKlonowski:
@pavel-stryber I plan to release v4.2.2 not later than end of this week (Friday).

PR #371, opened 2022-04-17
Fix: SeekBar on Android is rendering incorrectly in RTL mode

pavel-stryber:
This pull request #370   fixes SeekBar rendering in RTL mode and also adds ability to change layoutDirection depends on isRTL RN property.

Summary:
---------
1. Slider renders incorrectly on Android in rtl mode.
2. Currently slider changes it's layout direction based on device language. I want that it was relying on isRTL value of RN instead. It's important in case if we don't have translations for some rtl language and we have to use English as a default.

### DIFF
--- a/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/src/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -24,7 +24,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import javax.annotation.Nullable;
-
+import com.facebook.react.modules.i18nmanager.I18nUtil;
 /**
  * Slider that behaves more like the iOS one, for consistency.
  *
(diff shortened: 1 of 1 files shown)


BartoszKlonowski:
I am very happy to see this PR - thank you, @pavel-stryber!
I've allowed myself to edit the description so that it correctly links the PR to the issue you are fixing.

I will review this on Monday as I'm currently having Holiday break, thank you for your patiance.

BartoszKlonowski (commented):
Looks good from the code perspective, but I have a hard time debugging the fix:
I am trying this fix on Android emulator and even when whole app and system is RTL (after switching to RTL language) the Slider itself is still LTR oriented.

How should this fix working? Is it necessary to set `direction: "rtl"`?

It would be great if you could add a testing steps and maybe a working demo (some GIF or video) of your changes.

pavel-stryber:
> Looks good from the code perspective, but I have a hard time debugging the fix: I am trying this fix on Android emulator and even when whole app and system is RTL (after switching to RTL language) the Slider itself is still LTR oriented.
> 
> How should this fix working? Is it necessary to set `direction: "rtl"`?
> 
> It would be great if you could add a testing steps and maybe a working demo (some GIF or video) of your changes.

You need to make a few changes in example Android project.
- example/android/app/src/main/AndroidManifest.xml
add `android:supportsRtl="true"` to application
- example/android/app/src/main/java/com/example/MainActivity.java
add following code to onCreate method
```
I18nUtil sharedI18nUtilInstance = I18nUtil.getInstance();
sharedI18nUtilInstance.allowRTL(getApplicationContext(), true);
// if you run application with non-rtl system language, then enforce it with following
// sharedI18nUtilInstance.forceRTL(getApplicationContext(), true);
```

If it is still unclear I can share a branch with those changes.

BartoszKlonowski:
@pavel-stryber How about adding the required changes to the example app? From what I see they are all "allowing" to use RTL, not forcing it so it should be safe to have them.

pavel-stryber:
@BartoszKlonowski updated example project to allow RTL mode

BartoszKlonowski (approved):
Thank you!

PR #376, opened 2022-04-25
Update yarn.lock file

BartoszKlonowski:
This pull request updates the *yarn.lock* file by reinstalling the node_modules with clean lockfile.
All dependencies were refreshed according to the recent releases.

This pull request probably should fix at least some of the security/dependabot issues.


BartoszKlonowski:
Tested by running checks locally and on CI, also by running the example app (which uses the local "beta" version of Slider).

Issue #396, opened 2022-07-07
Repository major update


BartoszKlonowski:
The repository and the Slider itself are about to be updated.
The update will include:
* [x] leaving yarn workspaces
* [x] using TypeScript instead of Flow
* [ ] Use Swift and Kotlin for native implementation
* [x] Use the latest React & React Native

**Reasoning**:
I want to focus on the quality of Slider and it's becoming more and more difficult to do so with the legacy repo.
So I create this issue to keep the track on this huge update and keep notified all who is interested.

BartoszKlonowski:
Let's keep the Obj-C and Java as native languages for now. Currently a lot of implementation has been done and is currently in progress as well, so it would be a madness to change the language now.
And it should not block the further releases, so let's have it as it is, unless we get *really* tired or have way too much of a spare time and space.

PR #398, opened 2022-07-13
Update React & React Native versions

BartoszKlonowski:
This pull request is a part of #396 
It updates the React, React Native and React Native Windows versions to the latest according peer dependencies.
It also upgrades the other deps and tools for this PR to work like Jest, Babel, etc.


package/android/gradle.properties
```diff
@@ -1,4 +1,4 @@
-ReactNativeSlider_compileSdkVersion=29
+ReactNativeSlider_compileSdkVersion=31
 ReactNativeSlider_buildToolsVersion=28.0.3
-ReactNativeSlider_targetSdkVersion=27
-ReactNativeSlider_minSdkVersion=16
\ No newline at end of file
+ReactNativeSlider_targetSdkVersion=31
+ReactNativeSlider_minSdkVersion=21
```
BartoszKlonowski, line 4: I'm worried about the backward compatibility here,
while at the same time it feels odd leaving those at such low numbers.
Maybe we should keep the minimum intact, but update the target?

PR #401, opened 2022-07-24
Use GH Actions for CI

BartoszKlonowski:
This pull request configures the CI workflow based on GitHub Actions.

This setup uses mostly the latest Ubuntu runner and Macos, Windows for iOS, Windows examples builds.

Workflow is separated between:
* verifying Package sources (linting, testing, )
* verifying Example sources and building example apps
Building example apps also checks if the package sources are compilable

**Why not caching node_modules for Example jobs?**
Package sources are copied from package/ to example/ dir in a poistinstall step. If example's node_modules were cached, package sources would in fact be not checked whether they compile fine as they would be taken from cache, not from current commit.

**NOTES:**
* This pull request also removes unused now CI configuration for CircleCI, but the project is still turned on, which results in failing checks for this PR.
After merging, CircleCI should be disabled for this project.
* Additional job implementation for building new arch package should follow on new arch implementation branch.


.github/workflows/ReactNativeSlider-CI.yml
```diff
@@ -0,0 +1,161 @@
...
+on: [push]
+
+jobs:
+  install:
+    name: Install dependencies
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
thymikee, line 12:
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/ReactNativeSlider-CI.yml
```diff
@@ -0,0 +1,161 @@
...
+        uses: actions/setup-node@v2
+
+      - name: Display tools versions
+        run: npm --version;
+
+      - name: Restore all the packages
+        run: npm install
+
+      - name: Cache
+        uses: actions/cache@v2.1.6
```
thymikee, line 24: can we use v3?

thymikee (approved):

.github/workflows/ReactNativeSlider-CI.yml
```diff
@@ -0,0 +1,161 @@
...
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+
+      - name: Install node_modules on MacOS machine
+        run: npm install
+
+      - run: cd example/ios && xcode-select -p
```
deorst, line 128: Do we need this step or it's an output for debugging?

.github/workflows/ReactNativeSlider-CI.yml
```diff
@@ -0,0 +1,161 @@
...
+
+      - name: Install pod dependencies
+        run: cd example/ios && pod install --verbose
+
+      - name: Opening Simulator app
+        run: |
+          while ! open -Fn /Applications/Xcode_13.2.1.app/Contents/Developer/Applications/Simulator.app; do
```
deorst, line 135: I think we should refrain from hardcoding xcode version while using `macos-latest` runner, or hardcode runner version as well.
In order to open a simulator from command line we can try either `open -a Simulator.app` or [this](https://github.com/marketplace/actions/launch-ios-simulator) way of launching simulator in github actions.

deorst (approved):

.github/workflows/ReactNativeSlider-CI.yml
BartoszKlonowski: Debugging leftover, I'll remove that, thanks!

.github/workflows/ReactNativeSlider-CI.yml
BartoszKlonowski: We should, thanks!

PR #407, opened 2022-07-28
Fix: Pods-related error after using Pods from cache

BartoszKlonowski:
This pull request fixes issues with cached Pods that were unable to use after downloading from cache.
The fix is to reinstall downloaded Pods so that the Pods project will get reintegrated between different build and environments.

### DIFF
--- a/.github/workflows/ReactNativeSlider-CI.yml
+++ b/.github/workflows/ReactNativeSlider-CI.yml
@@ -105,23 +105,36 @@ jobs:
         id: cache-npm
         uses: actions/cache@v3
         env:
-          cache-name: cached-ios-deps
+          cache-name: cached-ios-npm-deps
         with:
-          path: |
-            example/node_modules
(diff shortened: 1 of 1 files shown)


okwasniewski:
I think it would be better to split caching between pods and node_modules to separate steps as @thymikee suggested.

BartoszKlonowski:
> I think it would be better to split caching between pods and node_modules to separate steps as @thymikee suggested.

Sure, but can you hand me a link to that? I think I did not yet see this.

BartoszKlonowski:
@okwasniewski @thymikee Before we merge this, I suggest rerunning the most recent workflow, because we need to check if build work fine with deps downloaded from cache, and we didn't really have that possibility in this PR yet.

thymikee (approved):
Let's verify the cache works and we're good 👍

okwasniewski (approved):
Looks good! 👍

Issue #395, opened 2022-07-07
onValueChanges is called on programmatic value changes on Android


eueddem:
## Environment
System:
    OS: macOS 11.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 1.06 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.16.0 - /var/folders/21/5xmhgtlx4bj0gthrmm8tm3zc0000gq/T/yarn--1657226700390-0.0757475079991885/node
    Yarn: 1.22.11 - /var/folders/21/5xmhgtlx4bj0gthrmm8tm3zc0000gq/T/yarn--1657226700390-0.0757475079991885/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.16.0/bin/npm
    Watchman: 2021.09.13.00 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.11.2 - /Users/eueddemkim-new/.rbenv/shims/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.2, iOS 15.2, macOS 12.1, tvOS 15.2, watchOS 8.3
    Android SDK:
      API Levels: 30, 31
      Build Tools: 30.0.2, 31.0.0
      Android NDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7784292
    Xcode: 13.2.1/13C100 - /usr/bin/xcodebuild
  Languages:
    Java: 1.8.0_292 - /usr/bin/javac
    Python: 2.7.16 - /usr/bin/python
  npmPackages:
    @react-native-community/cli: Not Found
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.4 => 0.63.4 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found

## Description
The doc says,
```onValueChanges: Callback continuously called while the user is dragging the slider.```
It works as expected on iOS device and simulators, but on Android devices(tested with Pixel 3 XL, Android 11), `onValueChanges` is called on every value changes, even from `setState`.

I'm on latest 4.2.4 version.
```"@react-native-community/slider": "^4.2.4",```

## Reproducible Demo
```tsx
const [state, setState] = useState(0);
const onValueChange = (value: number) => {
    console.log("onValueChange invoked", value);
};

.... (do state change after some time interval) ...

<Slider
[...]

PR #409, opened 2022-08-01
fix onValueChanges is called on programmatic value changes on Android

minhtc:
Summary:
---------

Revert changes from https://github.com/callstack/react-native-slider/pull/367 to fix https://github.com/callstack/react-native-slider/issues/395

What https://github.com/callstack/react-native-slider/pull/367 trying to fix is not an issue actually. The onValueChange should be called only when user is dragging the slide and it is matching with the behaviour on iOS

### DIFF
--- a/package/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
+++ b/package/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
@@ -108,7 +108,7 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
               new ReactSliderEvent(
                   seekbar.getId(),
                   ((ReactSlider)seekbar).toRealProgress(seekbar.getProgress()),
-                  !((ReactSlider)seekbar).isSliding()));
+                  false));
         }
       };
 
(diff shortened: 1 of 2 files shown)


BartoszKlonowski (changes requested):
Intuitively the `onValueChange` should be called also when user taps on Slider, as tapping also changes the value, I think it's the mistake of mine that we have the opposite written in the docs.
I think that instead of reverting the fix of one to fix another we should create an issue about the lack of `onValueChange` on iOS and work on a decent fix for the current issue keeping the backward compatibility.

minhtc:
> Intuitively the `onValueChange` should be called also when user taps on Slider, as tapping also changes the value, I think it's the mistake of mine that we have the opposite written in the docs.
> I think that instead of reverting the fix of one to fix another we should create an issue about the lack of `onValueChange` on iOS and work on a decent fix for the current issue keeping the backward compatibility.

If the `onValueChange` event is always be called, I think it should send with an extra flag to know if it is from dragging or tapping.
In my case, I need to check if the slider value is changed from dragging action to prevent some performance issue. If the slider value is changed by set the `value` prop, I do not need `onValueChange` because I already know the value.

jeffmon:
Sorry if this is not applicable here, but using the latest version `"@react-native-community/slider": "^4.2.4"`, `onValueChange` still fires automatically when the scrubber first starts up.

Dev-Grey (approved):

laurent22:
Also still happening in react-native-community/slider@4.4.0

PR #424, opened 2022-09-01
Use Typescript for package sources implementation

BartoszKlonowski:
This pull request introduces Typescript to the package implementation, instead of Flow.
This is done to package/ only, while example app is to be done for another PR - along with the example app update to v0.70 of react-native.

This implementation is part of #396

This PR updates also the CI to include the typechecking and modifies the sources compilation with Babel.


.github/workflows/ReactNativeSlider-CI.yml
```diff
@@ -39,6 +39,9 @@ jobs:
       - name: Run ESLint on the sources
         run: cd package && npx eslint src
 
+      - name: Run ESLint on the sources
```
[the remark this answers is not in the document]
BartoszKlonowski: Eh, copy-paste error. Thanks for catching this!

package/src/Slider.tsx
```diff
@@ -0,0 +1,273 @@
...
+  styles = StyleSheet.create({
+    slider: {
+      height: 40,
+    },
+  });
+} else {
+  styles = StyleSheet.create({
+    slider: {},
+  });
+}
```
okwasniewski, line 271: I think this can be done by using `Platform.select` or just a simple `Platform.OS` check without typing this to `any`.

package/src/RNCSliderNativeComponent.ts
```diff
@@ -0,0 +1,44 @@
...
+  tapToSeek?: WithDefault<boolean, false>;
+  maximumTrackImage?: ImageSource;
+  maximumTrackTintColor?: ColorValue;
+  maximumValue?: Float;
+  minimumTrackImage?: ImageSource;
+  minimumTrackTintColor?: ColorValue;
+  minimumValue?: Float;
+  onChange?: BubblingEventHandler<Event>;
+  onRNCSliderSlidingStart?: BubblingEventHandler<Event>;
+  onRNCSliderSlidingComplete?: BubblingEventHandler<Event>;
```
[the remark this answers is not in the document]
okwasniewski: Also just for reference typing from react-native slider implementation: https://github.dev/facebook/react-native/blob/73abcba40f305634efaafea8969cdfeb4982df0e/Libraries/Components/Slider/SliderNativeComponent.js#L51

package/src/Slider.tsx
```diff
@@ -0,0 +1,273 @@
...
+          : undefined
+      }
+      ref={forwardedRef}
+      style={style}
+      onChange={onValueChangeEvent}
+      onRNCSliderSlidingStart={onSlidingStartEvent}
+      onRNCSliderSlidingComplete={onSlidingCompleteEvent}
+      onRNCSliderValueChange={onValueChangeEvent}
+      enabled={!_disabled}
+      disabled={_disabled}
```
okwasniewski, line 241: This needs to be addressed, but in a separate PR.

package/typings/index.d.ts
```diff
@@ -158,9 +159,10 @@ export interface SliderProps extends SliderPropsIOS, SliderPropsAndroid, SliderP
...
 
 /**
-* A component used to select a single value from a range of values.
-*/
+ * A component used to select a single value from a range of values.
+ */
 declare class SliderComponent extends React.Component<SliderProps> {}
-declare const SliderBase: ReactNative.Constructor<ReactNative.NativeMethodsMixin> & typeof SliderComponent;
+declare const SliderBase: ReactNative.Constructor<ReactNative.NativeMethodsMixin> &
+  typeof SliderComponent;
```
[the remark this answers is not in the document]
BartoszKlonowski: Let's double check whether it's not a result of linter ran on this file.

package/src/Slider.tsx
BartoszKlonowski: Yes, already discussed while migrating to new architecture - let's create an issue for that: #426

package/src/RNCSliderNativeComponent.ts
```diff
@@ -0,0 +1,44 @@
...
+  fromUser?: boolean;
+}>;
+
+export interface NativeProps extends ViewProps {
+  accessibilityUnits?: string;
+  accessibilityIncrements?: ReadonlyArray<string>;
+  disabled?: WithDefault<boolean, false>;
+  enabled?: WithDefault<boolean, true>;
+  inverted?: WithDefault<boolean, false>;
+  vertical?: boolean;
```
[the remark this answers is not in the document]
BartoszKlonowski: No worries, let's have that as `false` by default 👍

package/typings/index.d.ts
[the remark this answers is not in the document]
BartoszKlonowski: Yes, it is the linter result, but now let's change only the 165-166 line, and keep the comment indendation, previously the comment was messed up, so it's now much better.

okwasniewski (approved):

.github/workflows/ReactNativeSlider-CI.yml
```diff
@@ -39,6 +39,9 @@ jobs:
       - name: Run ESLint on the sources
         run: cd package && npx eslint src
 
+      - name: Run TypeScript on the sources
+        run: cd package && npx tsc --noEmit
```
michalchudziak, line 43: Don't we want to run tsc on the example app as well?

package/src/__tests__/Slider.test.tsx
```diff
@@ -1,6 +1,5 @@
-/* @flow */
-
 import * as React from 'react';
+//@ts-ignore
```
michalchudziak, line 2: Why this ignore? there's https://www.npmjs.com/package/@types/react-test-renderer that we can use

package/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
michalchudziak, line 1: We might want to use https://www.npmjs.com/package/@tsconfig/react-native as a config foundation

michalchudziak (changes requested):

.github/workflows/ReactNativeSlider-CI.yml
BartoszKlonowski: I think we missed that during implementing #427

package/tsconfig.json
BartoszKlonowski: I don't think it's really needed - comparing our config to what comes with @tsconfig/react-native as recommended I can see that the only difference we have is the `target` being set to `ESNext` vs `ES6` in the proposed package. If we can avoid adding yet another dependency let's do so.

michalchudziak (approved):

PR #427, opened 2022-09-09
Upgrade example to 0.70 and use typescript

okwasniewski:
Summary:
---------

This pull requests introduces Typescript and  upgrades example app to react native 0.70. 

This is part of #396


example/SliderExample.js
```diff
@@ -1,240 +0,0 @@
...
-    },
-  },
-  {
-    title: 'Inverted slider direction',
-    render(): Element<any> {
-      return <SliderExample value={0.6} inverted />;
-    },
-  },
-  {
-    title: 'Vertical slider',
```
thymikee, line 220: this is not present in new example. I guess it's because windows platform. Do we have a plan to bring it back so we don't forget?

example/android/app/src/main/java/com/example/MainApplication.java
```diff
@@ -52,5 +55,37 @@ public void onCreate() {
     // If you opted-in for the New Architecture, we enable the TurboModule system
     ReactFeatureFlags.useTurboModules = BuildConfig.IS_NEW_ARCHITECTURE_ENABLED;
     SoLoader.init(this, /* native exopackage */ false);
+    initializeFlipper(this, getReactNativeHost().getReactInstanceManager());
```
BartoszKlonowski, line 58: Let's not bring the Flipper back - I removed it from example app as it was unused and I was trying to reduce the codebase.
Let's also remove the whole `initializeFlipper` method definition and `ReactNativeFlipper.java` file.

example/SliderExample.js
okwasniewski: Yes, @BartoszKlonowski temporarily disabled the build for windows for now, after the stable release for `react-native-windows@0.70` we will bring it back.

example/SliderExample.tsx
```diff
@@ -0,0 +1,188 @@
...
+    <View>
+      <SliderExample
+        {...props}
+        onSlidingComplete={value => {
+          setSlideStartingValue(value);
+          setSlideStartingCount(prev => prev + 1);
+        }}
+      />
+      <Text>
+        Completions: {slideStartingCount} Value: {slideStartingValue}
```
BartoszKlonowski, line 33: Should be `Starts: ` instead of completions.

BartoszKlonowski (commented):
We still have some JS files (even if they are config ones): babel.config.js, index.web.js, metro.config.js. Let's have them as TS.
Beside that, it looks good 👍

example/SliderExample.js
BartoszKlonowski: Let's not worry about it for now, in my opinion we should completely revisit the way we present examples.
Currently we divide them by platforms, but since first implemented some features changed and, for example iOS does not contain the disabled prop even if it's available on iOS.
I didn't want to put all those cleanings and refactoring and stuff into single PR to avoid making mess.
So @okwasniewski  let's continue with keeping all examples even if they are currently unavailable, and let's focus on reorganizing it in further PRs.

okwasniewski: I've added it back, since it's not called on other platforms, and just to not lose it.

okwasniewski:
> We still have some JS files (even if they are config ones): babel.config.js, index.web.js, metro.config.js. Let's have them as TS. Beside that, it looks good 👍

Thanks! I've added those files back, however due to those are config files they need to stay as `.js`. I've also commented out the config for windows.

BartoszKlonowski:
Ah, right! So let's have them as JSON - this should be doable.

BartoszKlonowski (approved):
Thank you!

thymikee (approved):
Let's create an issue for adding Windows back please and we're good 👍🏼

PR #432, opened 2022-09-26
feature: Add slide limit to iOS and Android

alfonsocj:
Summary:
---------
Feature to add slide limit. This will prevent sliding past the limit.

Sep-26-2022 12-36-29

- Added new `upperLimit` and `lowerLimit` props to Slider.
- Implemented `upperLimit` and `lowerLimit` on iOS and Android for fabric and paper.

Test Plan:
----------

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ To use this library you need to ensure you are using the correct version of Reac
 | `maximumValue` | Initial maximum value of the slider.<br/>Default value is 1. | number | No | |
 | `minimumTrackTintColor` | The color used for the track to the left of the button.<br/>Overrides the default blue gradient image on iOS. | [color](https://reactnative.dev/docs/colors) | No | |
 | `minimumValue` | Initial minimum value of the slider.<br/>Default value is 0. | number | No | |
+| `limit` | Slide limit. The user won't be able to slide past this limit.<br/>It is 0 when disabled. Default value is 0. | number | No | |
 | `onSlidingStart` | Callback that is called when the user picks up the slider.<br/>The initial value is passed as an argument to the callback handler. | function | No | |
 | `onSlidingComplete` | Callback that is called when the user releases the slider, regardless if the value has changed.<br/>The current value is passed as an argument to the callback handler. | function | No | |
 | `onValueChange` | Callback continuously called while the user is dragging the slider. | function | No | |
--- a/README.md
+++ b/README.md
@@ -89,7 +89,8 @@ To use this library you need to ensure you are using the correct version of Reac
 | `maximumValue` | Initial maximum value of the slider.<br/>Default value is 1. | number | No | |
 | `minimumTrackTintColor` | The color used for the track to the left of the button.<br/>Overrides the default blue gradient image on iOS. | [color](https://reactnative.dev/docs/colors) | No | |
 | `minimumValue` | Initial minimum value of the slider.<br/>Default value is 0. | number | No | |
-| `limit` | Slide limit. The user won't be able to slide past this limit.<br/>It is 0 when disabled. Default value is 0. | number | No | |
+| `lowerLimit` | Slide lower limit. The user won't be able to slide below this limit. | number | No | |
+| `upperLimit` | Slide upper limit. The user won't be able to slide above this limit. | number | No | |
 | `onSlidingStart` | Callback that is called when the user picks up the slider.<br/>The initial value is passed as an argument to the callback handler. | function | No | |
 | `onSlidingComplete` | Callback that is called when the user releases the slider, regardless if the value has changed.<br/>The current value is passed as an argument to the callback handler. | function | No | |
 | `onValueChange` | Callback continuously called while the user is dragging the slider. | function | No | |
--- a/README.md
+++ b/README.md
@@ -89,8 +89,8 @@ To use this library you need to ensure you are using the correct version of Reac
 | `maximumValue` | Initial maximum value of the slider.<br/>Default value is 1. | number | No | |
 | `minimumTrackTintColor` | The color used for the track to the left of the button.<br/>Overrides the default blue gradient image on iOS. | [color](https://reactnative.dev/docs/colors) | No | |
 | `minimumValue` | Initial minimum value of the slider.<br/>Default value is 0. | number | No | |
-| `lowerLimit` | Slide lower limit. The user won't be able to slide below this limit. | number | No | |
-| `upperLimit` | Slide upper limit. The user won't be able to slide above this limit. | number | No | |
+| `lowerLimit` | Slide lower limit. The user won't be able to slide below this limit. | number | No | Android, iOS |
+| `upperLimit` | Slide upper limit. The user won't be able to slide above this limit. | number | No | Android, iOS |
 | `onSlidingStart` | Callback that is called when the user picks up the slider.<br/>The initial value is passed as an argument to the callback handler. | function | No | |
 | `onSlidingComplete` | Callback that is called when the user releases the slider, regardless if the value has changed.<br/>The current value is passed as an argument to the callback handler. | function | No | |
 | `onValueChange` | Callback continuously called while the user is dragging the slider. | function | No | |
--- a/example/src/Examples.tsx
+++ b/example/src/Examples.tsx
@@ -3,7 +3,7 @@ import {Text, View, StyleSheet} from 'react-native';
 import Slider, {SliderProps} from '@react-native-community/slider';
 
 const SliderExample = (props: SliderProps) => {
-  const [value, setValue] = useState(0);
+  const [value, setValue] = useState(props.value ?? 0);
   return (
     <View>
       <Text style={styles.text}>{value && +value.toFixed(3)}</Text>
@@ -11,6 +11,7 @@ const SliderExample = (props: SliderProps) => {
         step={0.5}
         style={styles.slider}
         {...props}
+        value={value}
         onValueChange={setValue}
       />
     </View>
@@ -95,6 +96,14 @@ export const examples = [
     title: 'step: 0.25, tap to seek on iOS',
     render(): React.ReactElement {
       return <SliderExample step={0.25} tapToSeek={true} />;
+    },
+  },
+  {
+    title: 'Limit to 70',
+    render() {
+      return (
+        <SliderExample step={1} value={30} maximumValue={100} limit={70} />
+      );
     },
   },
   {
--- a/example/src/Examples.tsx
+++ b/example/src/Examples.tsx
@@ -99,10 +99,32 @@ export const examples = [
     },
   },
   {
-    title: 'Limit to 70',
+    title: 'Limit on positive values [30, 80]',
     render() {
       return (
-        <SliderExample step={1} value={30} maximumValue={100} limit={70} />
+        <SliderExample
+          step={1}
+          value={40}
+          minimumValue={0}
+          maximumValue={120}
+          lowerLimit={30}
+          upperLimit={80}
+        />
+      );
+    },
+  },
+  {
+    title: 'Limit on negative values [-70, -20]',
+    render() {
+      return (
+        <SliderExample
+          step={1}
+          value={-30}
+          minimumValue={-80}
+          maximumValue={0}
+          lowerLimit={-70}
+          upperLimit={-20}
+        />
       );
     },
   },
--- a/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -66,6 +66,12 @@ public class ReactSlider extends AppCompatSeekBar {
 
   private List<String> mAccessibilityIncrements;
 
+  /** Real limit value based on min and max values. This comes from props */
+  private double mRealLimit = 0;
+
+  /** Limit based on progress from 0..100 */
+  private int mLimit = 0;
+
   public ReactSlider(Context context, @Nullable AttributeSet attrs) {
     super(context, attrs);
     I18nUtil sharedI18nUtilInstance = I18nUtil.getInstance();
@@ -100,6 +106,15 @@ public class ReactSlider extends AppCompatSeekBar {
   /* package */ void setStep(double step) {
     mStep = step;
     updateAll();
+  }
+
+  /* package */ void setRealLimit(double value) {
+    mRealLimit = value;
+    updateAll();
+  }
+
+  int getLimit() {
+    return this.mLimit;
   }
 
   boolean isSliding() {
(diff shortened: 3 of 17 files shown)


example/SliderExample.tsx
```diff
@@ -3,14 +3,15 @@ import {Text, View, StyleSheet} from 'react-native';
 import Slider, {SliderProps} from '@react-native-community/slider';
 
 const SliderExample = (props: SliderProps) => {
-  const [value, setValue] = useState(0);
+  const [value, setValue] = useState(props.value || 0);
```
alfonsocj, line 6: I noticed that the SliderExample initial value was always 0, even though we pass pass it on a couple examples.

example/SliderExample.tsx
```diff
@@ -3,14 +3,15 @@ import {Text, View, StyleSheet} from 'react-native';
 import Slider, {SliderProps} from '@react-native-community/slider';
 
 const SliderExample = (props: SliderProps) => {
-  const [value, setValue] = useState(0);
+  const [value, setValue] = useState(props.value ?? 0);
```
alfonsocj, line 6: I noticed that the SliderExample initial value was always 0, even though we pass pass it on a couple examples.

alfonsocj:
@BartoszKlonowski Here's the PR as promised. Right now it only covers iOS and Android for both archs.

package/android/src/newarch/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -40,10 +40,17 @@ protected ViewManagerDelegate<ReactSlider> getDelegate() {
           new SeekBar.OnSeekBarChangeListener() {
             @Override
             public void onProgressChanged(SeekBar seekbar, int progress, boolean fromUser) {
+              ReactSlider slider = (ReactSlider)seekbar;
+
+              if(slider.getLimit() > 0 && progress > slider.getLimit()) {
```
BartoszKlonowski, line 45: What if `minimumValue = -100`, `maximumValue = -50` and `limit = -70`?
I'm afraid we have to consider the limit to be both positive and negative values including 0.
So to know whether Slider is limited we should introduce the internal field of Slider object: `boolean isLimited` which would be `false` as default and set to `true` if any value is passed to `limit` prop.

BartoszKlonowski (commented):
When it comes to the implementation: it's a clean and decent job 👍 
The only thing we need to think about is the API:
1. We should have limit set to any value, including negative and 0
For that let's have the boolean flag/field indicating whether optional `limit` prop has been set so we can take it into account when manipulating the value only `if (slider.isLimited() && progress > slider.getLimit()`.
2. Maybe we should consider both minimum and maximum limit?
If (as a user) I have a scenario where I can't set a value to max or just higher than my limit, then that is almost certain that I will as well have a scenario where I can't set a value lower than my `minimumLimit`.

package/android/src/newarch/java/com/reactnativecommunity/slider/ReactSliderManager.java
alfonsocj: Oops I forgot about negative values 😅. That makes sense. 
There's a problem with the `isLimited` approach. Since number is `0` by default when it gets to native, I couldn't find a way to identify if a value from JS is `undefined` or actually `0`. I'm changing my approach to set default limits to very small (for `lowerLimit`) and very big (for `upperLimit`) numbers. That way I don't need a boolean flag to limit progress.

alfonsocj:
@BartoszKlonowski I'm having a bug on Android. It looks like some race condition with `onProgressChanged` on the manager when the slider initial props are not yet set. It sets the slider value to `0`. This happens on the 5th slider in the example (on my branch), the value should be `40` but it quickly changes to `0`. Sometimes happens on the 6th slider too... Could you help me figure out what's wrong?

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -66,6 +66,18 @@ public class ReactSlider extends AppCompatSeekBar {
...
+  private double mRealLowerLimit = Long.MIN_VALUE;
+
+  /** Limit based on progress from 0..100 */
+  private int mLowerLimit = 0;
+
+  /** Real limit value based on min and max values. This comes from props */
+  private double mRealUpperLimit = Long.MAX_VALUE;
+
+  /** Limit based on progress from 0..100 */
+  private int mUpperLimit = 100;
```
alfonsocj, line 79: Setting lowest and highest limit possible for when the limits are not set

package/android/src/newarch/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -40,10 +40,22 @@ protected ViewManagerDelegate<ReactSlider> getDelegate() {
...
+
+              if(progress < slider.getLowerLimit()) {
+                progress = slider.getLowerLimit();
+                seekbar.setProgress(progress);
+              }
+
+              if(progress > slider.getUpperLimit()) {
+                progress = slider.getUpperLimit();
+                seekbar.setProgress(progress);
+              }
```
alfonsocj, line 53: The most important change for Android. It doesn't allow the seekbar to advance if the value is not within the limits

package/android/src/oldarch/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -28,11 +28,23 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> {
...
+
+              if(progress < slider.getLowerLimit()) {
+                progress = slider.getLowerLimit();
+                seekbar.setProgress(progress);
+              }
+
+              if(progress > slider.getUpperLimit()) {
+                progress = slider.getUpperLimit();
+                seekbar.setProgress(progress);
+              }
```
alfonsocj, line 41: Same as [the change above](https://github.com/callstack/react-native-slider/pull/432/files#diff-dc644f156dbc97ab8c8e17dcd152efba4e8b761399400b27eddba3a87640521bR43) but for the old arch

package/ios/RNCSliderComponentView.mm
```diff
@@ -118,6 +117,16 @@ - (void)RNCSendSliderEvent:(RNCSlider *)sender withContinuous:(BOOL)continuous i
...
+    if (value < sender.lowerLimit) {
+        value = sender.lowerLimit;
+        [sender setValue:value animated:NO];
+    }
+        
+    if (value > sender.upperLimit) {
+        value = sender.upperLimit;
+        [sender setValue:value animated:NO];
+    }
+
```
alfonsocj, line 129: Very similar to the Android implementation. Don't set the value when it's not within the limits.

package/ios/RNCSliderManager.m
```diff
@@ -87,6 +87,16 @@ static void RNCSendSliderEvent(RNCSlider *sender, BOOL continuous, BOOL isSlidin
...
+  if (value < sender.lowerLimit) {
+      value = sender.lowerLimit;
+      [sender setValue:value animated:NO];
+  }
+    
+  if (value > sender.upperLimit) {
+      value = sender.upperLimit;
+      [sender setValue:value animated:NO];
+  }
+
```
alfonsocj, line 99: Same as [the change above](https://github.com/callstack/react-native-slider/pull/432/files#diff-483b3787202f118ea009784458ee16d4020aab95f1d89b3994e0e1af07f33580R119) but for the old arch

package/src/Slider.tsx
```diff
@@ -221,9 +234,21 @@ const SliderComponent = (
...
+  const lowerLimit =
+    !!localProps.lowerLimit || localProps.lowerLimit === 0
+      ? localProps.lowerLimit
+      : LIMIT_MIN_VALUE;
+
+  const upperLimit =
+    !!localProps.upperLimit || localProps.upperLimit === 0
+      ? localProps.upperLimit
+      : LIMIT_MAX_VALUE;
+
```
alfonsocj, line 246: Setting defaults if the lower and upper limit props are not set.
The reason behind this is because for Android not setting them sends them as `0` to the native side which is a valid limit value.

alfonsocj:
@BartoszKlonowski I think it's ready for review 🙏

alfonsocj:
I updated the snapshots to include the `upperLimit` and `lowerLimit` props

alfonsocj:
All green! 🎉

alfonsocj:
@BartoszKlonowski, could you please review when you have a chance?

BartoszKlonowski:
@alfonsocj Yes, I will review this in upcoming days. Sorry for the delay!

package/ios/RNCSliderManager.m
```diff
@@ -87,6 +87,16 @@ static void RNCSendSliderEvent(RNCSlider *sender, BOOL continuous, BOOL isSlidin
 {
   float value = [sender discreteValue:sender.value];
 
+  if (value < sender.lowerLimit) {
+      value = sender.lowerLimit;
+      [sender setValue:value animated:NO];
+  }
+    
+  if (value > sender.upperLimit) {
```
BartoszKlonowski, line 95: Just like for Android - let's use `else if` here to avoid unnecessary calculations.

package/ios/RNCSliderComponentView.mm
```diff
@@ -118,6 +117,16 @@ - (void)RNCSendSliderEvent:(RNCSlider *)sender withContinuous:(BOOL)continuous i
 {
     float value = [sender discreteValue:sender.value];
     
+    if (value < sender.lowerLimit) {
+        value = sender.lowerLimit;
+        [sender setValue:value animated:NO];
+    }
+        
+    if (value > sender.upperLimit) {
```
BartoszKlonowski, line 125: Let's use `else if`.

BartoszKlonowski (commented):
Looks good when it comes to how it works and the implementation overall, but I do have a question and just a couple of non-major comments.
Thank you for taking the time to implement all this! 👍

alfonsocj:
@erikaannesmith I just tested your changes and it's looking good. Thanks!

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
alfonsocj: mUpperLimit range is from 0 to 100 (like a percentage). 100 being the rightmost value of the slider. It is set to 100 as default as this value has no effect on the slider.

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -186,9 +216,23 @@ private void updateAll() {
...
     setMax(getTotalSteps());
+    updateLowerLimit();
+    updateUpperLimit();
     updateValue();
   }
 
+  /** Update limit based on props limit, max and min */
+  private void updateLowerLimit() {
+    double limit = Math.max(mRealLowerLimit, mMinValue);
+    mLowerLimit = (int) Math.round((limit - mMinValue) / (mMaxValue - mMinValue) * getTotalSteps());
```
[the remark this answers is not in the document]
alfonsocj: The reason behind all these calculations is because the method `onProgressChanged` that the SeekBar change listener triggers gives us the SeekBar progress in the range of 0 to 20. So we translate all the Slider values to SeekBar values and be able to use them here https://github.com/callstack/react-native-slider/pull/432/files#diff-dc644f156dbc97ab8c8e17dcd152efba4e8b761399400b27eddba3a87640521bR45.

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
alfonsocj: Actually, I'm testing because this might not be true... `mUpperLimit` could be more than 100 if there are more than 100 total steps.

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -66,6 +66,18 @@ public class ReactSlider extends AppCompatSeekBar {
 
   private List<String> mAccessibilityIncrements;
 
+  /** Real limit value based on min and max values. This comes from props */
+  private double mRealLowerLimit = Long.MIN_VALUE;
+
+  /** Limit based on progress from 0..100 */
+  private int mLowerLimit = 0;
```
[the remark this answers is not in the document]
alfonsocj: The reason is that the SeekBar's progress is an int. On `onProgressChanged` we want to set the progress to the limit value if it goes above or below the limits. I'm now thinking if we need `mRealLowerLimit` and `mRealUpperLimit` to be double since it will never get there. I initially set them as `double` to match `mMinValue` and `mMaxValue` which are double.

Jay-CTSH:
Hi @alfonsocj, I'm kind of new to programming and the feature you implemented here is exactly something I need in my application. But I'm kind of confused since I've tried adding a upperLimit prop in the Slider component but the slider still doesn't stop at the upperLimit for some reason. What am I doing wrong?

```
<Slider
        minimumValue={0}
        value={0.5}
        maximumValue={1}
        upperLimit={0.7}
        onValueChange={value => rangeSetterSecond(value)}
      />
```

Stack Overflow Link: https://stackoverflow.com/questions/74741774/connecting-two-slider-components-to-act-as-one

alfonsocj:
Hello @Jay-CTSH, I think you are trying to use `upperLimit` or `lowerLimit` when that feature hasn't been merged yet. Hopefully it will be merged soon.

BartoszKlonowski:
Hello @Jay-CTSH, please make sure to use the changes implemented in this PR if you would like to test it, following what @alfonsocj said.

I will do my best to review and release the changes this week, so please expect v4.4.0 till 18.12.

README.md
```diff
@@ -89,6 +89,8 @@ To use this library you need to ensure you are using the correct version of Reac
 | `maximumValue` | Initial maximum value of the slider.<br/>Default value is 1. | number | No | |
 | `minimumTrackTintColor` | The color used for the track to the left of the button.<br/>Overrides the default blue gradient image on iOS. | [color](https://reactnative.dev/docs/colors) | No | |
 | `minimumValue` | Initial minimum value of the slider.<br/>Default value is 0. | number | No | |
+| `lowerLimit` | Slide lower limit. The user won't be able to slide below this limit. | number | No | |
+| `upperLimit` | Slide upper limit. The user won't be able to slide above this limit. | number | No | |
```
BartoszKlonowski, line 93: There's no Windows implemented for this feature yet, so as the Windows will not be supported let's add `iOS, Android` to the `Platform` column for both `lowerLimit` and `upperLimit`

example/src/Examples.tsx
```diff
@@ -3,14 +3,15 @@ import {Text, View, StyleSheet} from 'react-native';
...
-  const [value, setValue] = useState(0);
+  const [value, setValue] = useState(props.value ?? 0);
   return (
     <View>
       <Text style={styles.text}>{value && +value.toFixed(3)}</Text>
       <Slider
         step={0.5}
         style={styles.slider}
         {...props}
+        value={value}
```
BartoszKlonowski, line 14: What if we leave the `value` as default?

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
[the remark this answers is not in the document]
BartoszKlonowski: Thanks for explanation! 👍

BartoszKlonowski (approved):
Looks great, let's include platforms in the props table in README and let's ship it!
Thank you very much @alfonsocj, both for all the effort, figuring this out and for the patience along the way :)

example/src/Examples.tsx
alfonsocj: This would certainly work, but it feels wrong to me for the slider to not be a controlled component. What do you think?

PR #422, opened 2022-08-29
[Fabric] Refactor events, remove unused code

okwasniewski:
According to [this discussion](https://github.com/reactwg/react-native-new-architecture/discussions/70#discussioncomment-3429973), ShadowNodes are not used in the new architecture so I moved it back to old arch. 

Regarding events: 

> `RCTModernEventEmitter` is a transitional replacement for `RCTEventEmitter` that works with Fabric and non-Fabric renderers. `RCTEventEmitter` works with Fabric as well, but there are negative perf implications and it should be avoided.

Source: [react-native source code](https://github.dev/facebook/react-native/blob/e0a71fc7b5cb8c264000147099be5b2575931193/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/RCTModernEventEmitter.java#L7)

Test Plan:
----------
CI

---

This pull request fixes #412

PR #444, opened 2022-10-18
Replace dispatchModern() calls with getEventData()

### LINKED ISSUE #437: topChange null error while updating a property
## Environment

npmPackages:
    react: 18.2.0 => 18.2.0 
    react-native: 0.70.2 => 0.70.2 

## Description

After upgrade react-native-slider 4.31 to 4.3.2, display Slider crash on android.
No problem on iOS.
image

notjosh:
Summary:
---------

Over on https://github.com/callstack/react-native-slider/issues/437, we see issues on Android...sometimes. Long story short, Reanimated intercepts event dispatching, and it calls the method removed in #422. (there's some more detail in [this](https://github.com/callstack/react-native-slider/issues/437#issuecomment-1282577509) comment)

We can see in the stacktrace that `dispatchModern()` isn't being called when dispatched via Reanimated:

Screen Shot 2022-10-18 at 3 06 06 pm

The [react-native source](https://github.com/facebook/react-native/blob/f6cfcc3fc9e0b13d854f3c5295b67794507706c6/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/Event.java#L168-L172) describes the fix: don't override `dispatch()` and `dispatchModern()`, but simply override `getEventData()`. This PR does that, and seems to be working well!

(Sidebar: if we decide to keep the `dispatchModern()` override, we should update the implementation to use the actual modern dispatcher instead of the old dispatcher. See [here](https://github.com/facebook/react-native/blob/f6cfcc3fc9e0b13d854f3c5295b67794507706c6/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/Event.java#L211-L227) for how that should look.)

This pull request fixes #437 

Test Plan:
----------

I have a repro on [this](https://github.com/notjosh/react-native-slider/tree/crashy) branch (it crashes on boot, no repro steps necessary), and we can use [this](https://github.com/notjosh/react-native-slider/tree/fixy) branch to verify that the change works with Reanimated.

Note: It's possible that other libs than Reanimated have other ways of calling this, so there might be a side effect to implementing `getEventData()`, but I _think_ it should be pretty safe.

### DIFF
--- a/package/android/src/main/java/com/reactnativecommunity/slider/ReactSliderEvent.java
+++ b/package/android/src/main/java/com/reactnativecommunity/slider/ReactSliderEvent.java
@@ -7,10 +7,10 @@
 
 package com.reactnativecommunity.slider;
 
+import androidx.annotation.Nullable;
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.WritableMap;
 import com.facebook.react.uimanager.events.Event;
-import com.facebook.react.uimanager.events.RCTModernEventEmitter;
 
 /**
  * Event emitted by a ReactSliderManager when user changes slider position.
@@ -47,9 +47,10 @@ public class ReactSliderEvent extends Event<ReactSliderEvent> {
     return 0;
   }
 
+  @Nullable
   @Override
-  public void dispatchModern(RCTModernEventEmitter rctEventEmitter) {
-    rctEventEmitter.receiveEvent(getViewTag(), getEventName(), serializeEventData());
+  protected WritableMap getEventData() {
+    return serializeEventData();
   }
 
   private WritableMap serializeEventData() {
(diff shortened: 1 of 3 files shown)


BartoszKlonowski:
Thank you, @notjosh! 👍 
Please give me some time to review this and merge. In the meantime, please rebase the PR on top of the `main` branch to apply the #445

notjosh:
Ah, whoops, I realised the diff was wrong - when editing the example in Android Studio, it's only writing into `node_modules` instead of the main project 🤦

I've updated the imports to actually include `Nullable`, which builds locally via the same command as the CI step...so hopefully CI passes now!

Willham12:
lgtm!

BartoszKlonowski (approved):
Looks good, thank you for this contribution 👍

PR #448, opened 2022-10-22
Fix @types being present as a prod dependency

liamjones:
Summary:
---------

@types/react-test-renderer was added for a test but was added as a `dependency` rather than a `devDependency`. This results in it being a transitive prod dependency for any project using react-native-slider.

Test Plan:
----------

npm run lint and npm run test ran.
test still passes. lint complains no more than it did before the change was made against main.


BartoszKlonowski (approved):
Hello @liamjones, Thank you very much for fixing this! 👍

Issue #426, opened 2022-09-08
Remove redundant `enabled` property


BartoszKlonowski:
The Slider implementation on RN side has both `enabled` `disabled` used, which is redundant and can lead to mistakes.
`enabled` property should be removed in favor of `disabled` which matches the name from docs.

This needs to be addressed, but in a separate PR.

_Originally posted by @okwasniewski in https://github.com/callstack/react-native-slider/pull/424#discussion_r966294006_

PR #449, opened 2022-10-28
fix: remove redundant "enabled" property

slavikdenis:
Summary:
---------

Removing redundant "enabled" property as discussed in https://github.com/callstack/react-native-slider/pull/424#discussion_r966294006 and documented in https://github.com/callstack/react-native-slider/issues/426

Fixed some instructions in README as well:
- iOS: "`npm run ios` in `example/ios` folder" instead of "`npm run example-ios`" <-- the other command runs `pod-install` which installs pods with old arch
- Android: adding note that after changing settings to new architecture, first run fails for some reason and other ones are running fine 🤷 

Test Plan:
----------

Enabled "disabled" example for all platforms, so run the examples and test it out :)

iOS:
simulator_screenshot_0E3EFDC2-4119-4CD8-B8F7-7FC738E92C2E

Android:
Screenshot_1666969491

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ You can also do this manually by:
 In order to use the new architecture some extra steps are needed. 
 #### iOS 
 - Install pods with new arch flag inside `example/ios` folder: `RCT_NEW_ARCH_ENABLED=1 pod install`
-- Run `npm run ios` in `example/ios` folder
+- Run `npm run example-ios`
 
 #### Android 
 - Set `newArchEnabled` to true inside `example/android/gradle.properties`
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ In order to use the new architecture some extra steps are needed.
 
 #### Android 
 - Set `newArchEnabled` to true inside `example/android/gradle.properties`
-- Run `npm run example-android` _(first run might fail, try to run it again)_
+- Run `npm run example-android`
 
 <details>
 <summary>
--- a/README.md
+++ b/README.md
@@ -146,11 +146,11 @@ You can also do this manually by:
 In order to use the new architecture some extra steps are needed. 
 #### iOS 
 - Install pods with new arch flag inside `example/ios` folder: `RCT_NEW_ARCH_ENABLED=1 pod install`
-- Run `npm run example-ios`
+- Run `npm run ios` in `example/ios` folder
 
 #### Android 
 - Set `newArchEnabled` to true inside `example/android/gradle.properties`
-- Run `npm run example-android`
+- Run `npm run example-android` (first run might fail, try to run it again)
 
 <details>
 <summary>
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -146,11 +146,11 @@ You can also do this manually by:
 In order to use the new architecture some extra steps are needed. 
 #### iOS 
 - Install pods with new arch flag inside `example/ios` folder: `RCT_NEW_ARCH_ENABLED=1 pod install`
-- Run `npm run example-ios`
+- Run `npm run ios` in `example/ios` folder
 
 #### Android 
 - Set `newArchEnabled` to true inside `example/android/gradle.properties`
-- Run `npm run example-android`
+- Run `npm run example-android` _(first run might fail, try to run it again)_
```
BartoszKlonowski, line 153: This is worrying... What do we mean here by "first run"? Do we mean the build? Or the app launch?
We don't have anything like that reported and I didn't observe it yet neither in my work with example app, neither in CI builds.

Let me double check it so we make 100% sure it's not a local issue of yours that we put to official docs - for that can you create also the issue in repo so I can check it in repro app?

slavikdenis: This happened in this scenario:
- clone project
- run install
- run android on the old arch
- change settings to the new arch
- run android on the new arch (without cleaning android <-- might be the problem)

It might have failed on some of the downloads that were going on, but not sure. If you prefer, I can remove the note in the README. Just put it there so people might try it out once more before creating an issue

BartoszKlonowski: @slavikdenis Thanks for explaining 👍 
Hm, so if this issue appears without cleaning, then I'd prefer to have a note in README that the clean of Android build is suggested when building new arch on an old arch.
Yes, let's remove that from this PR - I think that no matter what kind of note we decide to put in README it's always better to separate such changes between two PRs.
But thank you for noticing this and letting me know, it's definitely something worth revisiting.

slavikdenis: Removed via https://github.com/callstack/react-native-slider/pull/449/commits/8303f8af7a29b4ba8b48d85d622c4a91f6f86e8b

Let's see if it occurs to anyone else 👍

README.md
```diff
@@ -146,7 +146,7 @@ You can also do this manually by:
 In order to use the new architecture some extra steps are needed. 
 #### iOS 
 - Install pods with new arch flag inside `example/ios` folder: `RCT_NEW_ARCH_ENABLED=1 pod install`
-- Run `npm run example-ios`
```
BartoszKlonowski, line 149: Let's remove that change.
Having this like so is intentional - it guides how to launch the example from project's root where the README is.
We can add the proposed step as additional, alternative like: "`or run npm run ios`(...)" but still it should be done in separate PR. This PR is about redundant `enabled` property.

slavikdenis: Done ✔️

BartoszKlonowski (approved):
Looks good, Thank you! 👍

PR #468, opened 2022-12-22
Updated the Touch Height for the ios device.

kaushallalo:
Summary:
---------

Test Plan:
----------
I needed to customize the height of the touch area for the ios so i have made changes to the file.

### DIFF
--- a/package/src/Slider.tsx
+++ b/package/src/Slider.tsx
@@ -285,7 +285,7 @@ SliderWithRef.defaultProps = {
 };
 
 let styles = StyleSheet.create(
-  Platform.OS === 'ios' ? {slider: {height: 40}} : {slider: {}},
+  Platform.OS === 'ios' ? {slider: {}} : {slider: {}},
 );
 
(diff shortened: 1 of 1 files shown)


kaushallalo:
Please review my changes and accept my PR

package/src/Slider.tsx
```diff
@@ -285,7 +285,7 @@ SliderWithRef.defaultProps = {
 };
 
 let styles = StyleSheet.create(
-  Platform.OS === 'ios' ? {slider: {height: 40}} : {slider: {}},
+  Platform.OS === 'ios' ? {slider: {}} : {slider: {}},
```
BartoszKlonowski, line 288: With removing the `height: 40` style, this line does actually nothing, and could even be simplified into:
```suggestion
  {slider: {}}
```

BartoszKlonowski (commented):
I'm trying to understand the reason of this change - does the default style created for iOS conflicts with the custom height for you?

PR #472, opened 2022-12-27
fix(web): invalidate position on ancestor scroll

### LINKED ISSUE #471: Slider doesn't work in a horizontal scrolling container on web (e.g. tab view)
## Environment

N/A, reproduces on Snack

## Description

If a slider is rendered inside a container that can scroll horizontally, the [cached container X position](https://github.com/callstack/react-native-slider/blob/main/package/src/RNCSliderNativeComponent.web.tsx#L213-L219) is not cleared on scroll, which throws off the calculation in [`getValueFromNativeEvent`](https://github.com/callstack/react-native-slider/blob/main/package/src/RNCSliderNativeComponent.web.tsx#L221-L242) and makes the slider unusable.

One implication of this - and where it bit me - is that the component can't be used reliably inside a tab view that relies on scrolling (e.g. `@react-navigation/material-top-tabs`). This manifests as sliders being completely unresponsive on all tabs except the one that was initially opened during the tab view's first render.

## Reproducible Demo

https://snack.expo.dev/@motiz88/slider-horizontal-slide-bug

[](https://user-images.githubusercontent.com/2246565/209706806-e6c56f08-1b5a-4e6a-9e52-d674ccaa8cd1.mov)

motiz88:
Summary:
---------

Fixes #471.

As there is no native way to listen for changes to `getBoundingClientRect()`, the web implementation of Slider must cache and try to correctly invalidate the measurements. The code already does this on layout and when the viewport is resized, but not when the slider's bounding rect changes due to scrolling. This PR registers a capturing `scroll` event handler on `document` and invalidates the cache when an ancestor of the slider scrolls. This is probably not airtight (see this [StackOverflow discussion](https://stackoverflow.com/questions/40251082/an-event-or-observer-for-changes-to-getboundingclientrect)) but is AFAICT correct.

For performance, it may be worth experimenting with unregistering the `scroll` (and `resize`) listener once the cache has been invalidated - this would avoid calling O(sliders on page) handlers on every frame while scrolling. This is a potentially more invasive change to Slider so I am publishing the PR without it.

Test Plan:
----------

Added an example to `example-web` and tested it manually (see video).

https://user-images.githubusercontent.com/2246565/209722872-273ba577-6ae8-44b3-957c-6455fbc93827.mov

### DIFF
--- a/package/src/RNCSliderNativeComponent.web.tsx
+++ b/package/src/RNCSliderNativeComponent.web.tsx
@@ -69,7 +69,7 @@ const RCTSliderWebComponent = React.forwardRef(
     const containerSize = React.useRef({width: 0, height: 0});
     const containerPositionX = React.useRef(0);
     const containerRef = forwardedRef || React.createRef();
-    const hasBeenResized = React.useRef(false);
+    const containerPositionInvalidated = React.useRef(false);
     const [value, setValue] = React.useState(initialValue || minimumValue);
     const lastInitialValue = React.useRef<number>();
 
@@ -145,18 +145,36 @@ const RCTSliderWebComponent = React.forwardRef(
       }
     }, [initialValue, updateValue, animatedValue]);
 
-    const onResize = () => {
-      hasBeenResized.current = true;
-    };
     React.useEffect(() => {
+      const invalidateContainerPosition = () => {
+        containerPositionInvalidated.current = true;
+      };
+      const onDocumentScroll = (e: Event) => {
+        if (
+          // Avoid all other checks if already invalidated
+          !containerPositionInvalidated.current &&
+          containerRef.current &&
+          // @ts-ignore
+          e.target.contains(containerRef.current)
+        ) {
+          invalidateContainerPosition();
+        }
+      };
       //@ts-ignore
-      window.addEventListener('resize', onResize);
+      window.addEventListener('resize', invalidateContainerPosition);
+      //@ts-ignore
+      document.addEventListener('scroll', onDocumentScroll, {capture: true});
 
       return () => {
         //@ts-ignore
-        window.removeEventListener('resize', onResize);
+        window.removeEventListener('resize', invalidateContainerPosition);
+
+        //@ts-ignore
+        document.removeEventListener('scroll', onDocumentScroll, {
+          capture: true,
+        });
       };
-    }, []);
+    }, [containerRef]);
 
     const containerStyle = StyleSheet.compose(
       {
@@ -221,8 +239,8 @@ const RCTSliderWebComponent = React.forwardRef(
     const getValueFromNativeEvent = (pageX: number) => {
       const {width = 1} = containerSize.current;
 
-      if (hasBeenResized.current) {
-        hasBeenResized.current = false;
+      if (containerPositionInvalidated.current) {
+        containerPositionInvalidated.current = false;
         updateContainerPositionX();
       }
       const containerX = containerPositionX.current;
--- a/package/src/RNCSliderNativeComponent.web.tsx
+++ b/package/src/RNCSliderNativeComponent.web.tsx
@@ -150,9 +150,9 @@ const RCTSliderWebComponent = React.forwardRef(
         containerPositionInvalidated.current = true;
       };
       const onDocumentScroll = (e: Event) => {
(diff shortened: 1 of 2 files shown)


package/src/RNCSliderNativeComponent.web.tsx
```diff
@@ -145,18 +145,36 @@ const RCTSliderWebComponent = React.forwardRef(
...
+          containerRef.current &&
+          // @ts-ignore
+          e.target.contains(containerRef.current)
+        ) {
+          invalidateContainerPosition();
+        }
+      };
       //@ts-ignore
-      window.addEventListener('resize', onResize);
+      window.addEventListener('resize', invalidateContainerPosition);
```
jspizziri, line 164: @motiz88 you should probably subscribe to resize events the react-native way:

https://necolas.github.io/react-native-web/docs/use-window-dimensions/

package/src/RNCSliderNativeComponent.web.tsx
```diff
@@ -145,18 +145,36 @@ const RCTSliderWebComponent = React.forwardRef(
...
+      //@ts-ignore
+      document.addEventListener('scroll', onDocumentScroll, {capture: true});
```
jspizziri, line 166: If there's no way to do this in a more react-native idiomatic way, my opinion would be to do something like the following at the top of the file:

```ts
// @ts-ignore
const { addEventListener, removeEventListener } = document;

...
const RCTSliderWebComponent = React.forwardRef(
...
);

```

That would reduce the overall number of `@ts-ignore` comments in the code, which aren't ideal.

Perhaps it would be even better to create a unique hook (e.g. `useOnDocumentScroll` that receives a callback or something), that abstracts this functionality.

package/src/RNCSliderNativeComponent.web.tsx
motiz88: @jspizziri
1. This use of the `resize` event was pre-existing and I had no intention of changing it as part of this bugfix. I merely renamed `onResize` to `invalidateContainerPosition` to better communicate what it does (and reuse it in the `onScroll` handler).
2. IMO, given this is platform-specific code, use of DOM primitives is completely fine. It's also not directly interchangeable with `useWindowDimensions` (which would tie us to the React component lifecycle in a way that might not be the most correct/efficient for this use case)

package/src/RNCSliderNativeComponent.web.tsx
motiz88: > That would reduce the overall number of @ts-ignore comments in the code, which aren't ideal.

I think the `@ts-ignore` issue (which, again, is pre-existing here) would be best solved by configuring TypeScript to properly typecheck uses of DOM APIs in `.web.tsx` files. This might be solved by adding `/// <reference lib="dom" />` to the top of the file, but I haven't tried it.

motiz88: @jspizziri I've put up #475 to deal with the `@ts-ignore`s comprehensively.

package/src/RNCSliderNativeComponent.web.tsx
```diff
@@ -145,18 +145,36 @@ const RCTSliderWebComponent = React.forwardRef(
...
+        containerPositionInvalidated.current = true;
+      };
+      const onDocumentScroll = (e: Event) => {
+        if (
+          // Avoid all other checks if already invalidated
+          !containerPositionInvalidated.current &&
+          containerRef.current &&
+          // @ts-ignore
+          e.target.contains(containerRef.current)
+        ) {
```
BartoszKlonowski, line 159: I'm not sure if linting will accept the formatting, but anyway I often prefer introducing a properly named const instead of a comment:
```suggestion
        const isAlreadyInvalidated = !containerPositionInvalidated.current;
        if (
          isAlreadyInvalidated &&
          containerRef.current &&
          // @ts-ignore
          e.target.contains(containerRef.current)
        ) {
```
It's just a suggestion - accept only if you feel like it 👍

BartoszKlonowski (approved):
To me it looks very good, I'm happy to accept that. Also thanks so much for adding another example showing this specific scenario! 👍 

@jspizziri Would it be possible for you to let us know if @motiz88's answers address your remarks? I don't want to merge something that still has some potentially open discussions.

PR #482, opened 2023-01-14
Changed `!!` with `Boolean` to make the condition more readable

faizqadri234:
Summary:
---------
Changed `!!` with `Boolean` to make the condition more readable

Test Plan:
----------
My change has no effect on the logic.

### DIFF
--- a/package/src/Slider.tsx
+++ b/package/src/Slider.tsx
@@ -235,12 +235,12 @@ const SliderComponent = (
     : null;
 
   const lowerLimit =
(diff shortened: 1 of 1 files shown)


BartoszKlonowski (commented):
Hello @faizqadri234, thanks for contribution!
I don't think we should change that - the `!!` operator is (in my experience) much more widely used among developers and for those it may also be as much readable as `Boolean()`. I really appreciate the will to contribute and I encourage to give it a try with some of reported issues, but having that said I'd rather avoid changing that.

PR #486, opened 2023-01-24
fix(types): utility type removed from RN 0.71

alpha0010:
Summary:
---------

Updating to react native 0.71, TypeScript errors on this library. The utility type `Constructor` is private in latest typings. Using this export degrades to `any` type.

`JSX element class does not support attributes because it does not have a 'props' property.`

Test Plan:
----------

`npx tsc` runs clean after making this change on a RN 0.71 app.

### DIFF
--- a/package/typings/index.d.ts
+++ b/package/typings/index.d.ts
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import * as ReactNative from 'react-native';
 
+type Constructor<T> = new (...args: any[]) => T;
+
 type SliderReferenceType =
   | (React.MutableRefObject<SliderRef> & React.LegacyRef<Slider>)
   | undefined;
@@ -172,7 +174,7 @@ export interface SliderProps
  * A component used to select a single value from a range of values.
  */
 declare class SliderComponent extends React.Component<SliderProps> {}
-declare const SliderBase: ReactNative.Constructor<ReactNative.NativeMethods> &
+declare const SliderBase: Constructor<ReactNative.NativeMethods> &
   typeof SliderComponent;
 export default class Slider extends SliderBase {}
 export type SliderIOS = Slider;


package/typings/index.d.ts
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import * as ReactNative from 'react-native';
 
+type Constructor<T> = new (...args: any[]) => T;
```
BartoszKlonowski, line 4: Hello @alpha0010, thank you for contributing!
I do see the general reasoning behind this, but I can't figure out the need to change this for this specific package that is currently running on react-native v0.69.1.
Are you able to elaborate on why changing this here? Or maybe does the original implementation currently affects your app that is running on version you mentioned (v0.71)?

xKRUTOIx: Hi! 

I think I know what's the reason behind this PR. 

After upgrading to RN 0.71 my IDE has started highlighting the same issue as @alpha0010 had mentioned.

Basically it looks like this:
image


After applying the suggested change it goes away.

alpha0010: RN 0.71 bundles TS types https://reactnative.dev/blog/2023/01/03/typescript-first#declarations-shipped-with-react-native , which hides this type from general export https://github.com/facebook/react-native/blob/v0.71.1/types/private/Utilities.d.ts .

LogicNg: Same problem here

BartoszKlonowski: Oh yeah, so indeed:
> Or maybe does the original implementation currently affects your app that is running on version you mentioned (v0.71)?

Is the case here.
Alright!

BartoszKlonowski (approved):
OK, let's have that! Thank you!

package/typings/index.d.ts
allen-hsu: Same issue

PR #519, opened 2023-07-03
feat: Add support for RN 0.72

mrousavy:
Adds `namespace` to `build.gradle` (this is required by Gradle 8, which RN 0.72 uses)

### DIFF
--- a/package/android/build.gradle
+++ b/package/android/build.gradle
@@ -31,6 +31,7 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+  namespace 'com.reactnativecommunity.slider'
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
   buildToolsVersion getExtOrDefault('buildToolsVersion')
 


gabrielbarth (approved):

package/android/build.gradle
```diff
@@ -31,6 +31,7 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+  namespace 'com.reactnativecommunity.slider'
```
gabrieldonadel, line 34: Might be better to check for the agp version
```suggestion
    def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
    if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
        namespace "com.reactnativecommunity.slider"
    }
```

satya164: It's only available from 7.3+, not 7, so this check will fail if user is on an older version.

BartoszKlonowski:
Superseded by #553

PR #553, opened 2023-11-10
fix(android): android gradle plugin 8 compatibility

mikehardy:
Summary:
---------

This is the minimum required change for this module to work on react-native 0.73 which includes android gradle plugin 8

This module does not just require namespace it also requires the buildConfig feature turned on, as discovered in actual use in a work app as I tested AGP8+ to fix up modules I maintain and modules I depend on

* Supercedes #519

It is similar to changes I needed to do as react-native-firebase maintainer --> https://github.com/invertase/react-native-firebase/commit/b52d0ce6723c077190618641ce0f33ced9fd4090

## Test Plan

With apologies, you have to alter an app that integrates this module to use android gradle plugin 8, it's difficult to do that in repos I'm proposing these changes in because bumping to android gradle plugin 8 requries a large amount of transitive dependency changes in CI

I have integrated this in an app and tested it, and done similar work as maintainer of react-native-firebase, react-native-netinfo and react-native-device-info, and I'm now pushing these out to the repos

Cheers

### DIFF
--- a/package/android/build.gradle
+++ b/package/android/build.gradle
@@ -34,8 +34,8 @@ android {
 
   def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
   // Check AGP version for backward compatibility w/react-native versions still on gradle plugin 6
-  def major = parsed[0].toInteger()
-  def minor = parsed[1].toInteger()
+  def major = agpVersion[0].toInteger()
+  def minor = agpVersion[1].toInteger()
   if ((major == 7 && minor >= 3) || major >= 8) {
     namespace "com.reactnativecommunity.slider"
     buildFeatures {
--- a/package/android/build.gradle
+++ b/package/android/build.gradle
@@ -32,7 +32,7 @@ def getExtOrIntegerDefault(name) {
 
 android {
 
-  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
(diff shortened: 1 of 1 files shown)


BartoszKlonowski (approved):
Thank you for the contribution, @mikehardy 👍 
I've allowed myself to add a suggestion regarding the specific v7.3 instead of just v7, to make sure we do not add the namespace for versions in between.
Once it's CI-checked, I'll merge your PR.

package/android/build.gradle
```diff
@@ -31,6 +31,18 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+  // Check AGP version for backward compatibility w/react-native versions still on gradle plugin 6
+  def major = parsed[0].toInteger()
+  def minor = parsed[1].toInteger()
```
BartoszKlonowski, line 38:
```suggestion
  def major = agpVersion[0].toInteger()
  def minor = agpVersion[1].toInteger()
```

package/android/build.gradle
```diff
@@ -31,6 +31,18 @@ def getExtOrIntegerDefault(name) {
 }
 
 android {
+
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
```
BartoszKlonowski, line 35:
```suggestion
  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')
```

Issue #567, opened 2023-12-27
Issue with Thumb Positioning in Web Implementation


landabaso:
**Environment:** Web implementation of `react-native-slider`.

**Problem Summary:** There is an issue with the thumb positioning in the web implementation. This problem is noticeable when using sliders without the `step` property (such as the examples towards the end of the `example-web` folder, f.ex.: the one titled "trackImage").

**Steps to Reproduce:**
1. Run the `example-web` provided in the `react-native-slider` repository.
2. Navigate to one of the slider examples without a `step` value (e.g., the one titled "trackImage").
3. Position the slider thumb at one of the extremes (either the very left or right end).
4. Click on the center of the thumb.

**Expected Behavior:** When clicking on the center of the thumb, especially at the extremes, the thumb position should accurately represent the slider value and remain stationary if the click doesn't intend to change the value.

**Actual Behavior:** Upon clicking on the center of the thumb when it's positioned at one of the extremes, the thumb slightly moves towards the center and a new value is returned. It's subtle but noticeable. This issue appears to stem from the way the thumb's container is sized, which includes the thumb circle itself, leading to an offset in value representation.

**Additional Information:** This issue affects the usability of the slider. The problem seems to be with the calculation of the thumb's position in relation to the slider's container size, which does not account for the size of the thumb circle itself.

**Note:** I will be submitting a PR shortly for review and further discussion.

PR #568, opened 2023-12-27
Fix Thumb Positioning on Web (#567)

landabaso:
This PR addresses the issue #567 where the slider thumb on web implementations would incorrectly position itself. It's specially noticeable when clicked at the extremes.

Changes:
- Adjusted the calculation of the thumb's position in `getValueFromNativeEvent` to consider the thumb size.
- Ensured the thumb remains within the bounds of the slider track.

The fix was tested with the `example-web`. The thumb now accurately reflects the slider value, and its position remains stable when clicked at the extremes.

Looking forward to feedback or suggestions.


BartoszKlonowski (approved):
This looks alright, thank you!

Issue #155, opened 2020-02-19
Slider does not respect the step value on when using with keyboard


maximilize:
## Environment

```
System:
    OS: Linux 4.19 Fedora 30 (Thirty) 30 (Thirty)
    CPU: (4) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
    Memory: 330.60 MB / 9.36 GB
    Shell: 5.0.11 - /bin/bash
  Binaries:
    Node: 13.8.0 - ~/.local/node-v13.8.0-linux-x64/bin/node
    npm: 6.13.6 - ~/.local/node-v13.8.0-linux-x64/bin/npm
  npmPackages:
    @react-native-community/cli: ^4.1.0 => 4.1.0 
    react: ^16.9.0 => 16.12.0 
    react-native: ^0.61.5 => 0.61.5
```

## Description

When using the slider with for example a remote control or keyboard, the `step` property is ignored. This does not allow us to do fine grained sliding. On a full width slider there are maybe 16 steps.

The goal should be on a minimum value of 0, and a maximum value of 1000, and a `step` value of 1, 1000 steps.

michalchudziak:
Hey, @maximilize  thanks for reporting! I'd appreciate any PRs fixing this issue :)

Sharcoux:
@michalchudziak should be working on Web. https://github.com/react-native-community/react-native-slider/issues/169

Sharcoux:
Could you confirm that the problem disappear when using @sharcoux/slider from npm ?

BartoszKlonowski:
Hello @maximilize!
I couldn't reproduce the same issue when trying it on the emulator.
Could you check if the same issue is solved on your side using the latest version?

vafada:
@BartoszKlonowski this is still an issue

here's a video:
![bug](https://github.com/callstack/react-native-slider/assets/285941/ac38b630-8dbf-40f9-bc97-d539cf9bf3f8)

The issue is in this line:

https://github.com/callstack/react-native-slider/blob/main/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java#L215-L218

because of `DEFAULT_TOTAL_STEPS = 128;`

on load `mStepCalculated = 0.234375`

so `getTotalSteps` will be `128` at initial load

calling `setMax(128)` changes the internal `mKeyProgressIncrement` to `6`

see: 

[...]

PR #572, opened 2024-01-31
fix keyboard progress increment getting stuck at 6

vafada:
Summary:
---------

This fixes https://github.com/callstack/react-native-slider/issues/155

see https://github.com/callstack/react-native-slider/issues/155#issuecomment-1918183040 for details of the issue

The gist of it, on initial load `ReactSlider` calls

`setMax(128)`

which calls the `AbsSeekBar.java`'s `setMax()`

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/widget/AbsSeekBar.java;l=586-595;drc=c8c729556d69aa27fff24e5412140563fc1527af

and internally updates `keyProgressIncrement` to `6`

that causes keyboard input to jump 6 steps:

https://android.googlesource.com/platform/frameworks/base/+/711efaa0297f845c60a1de0acce680493f90bf02/core/java/android/widget/AbsSeekBar.java#1053

after setting  `ReactSlider.setMaxValue`, if `max` is not high enough, the  `keyProgressIncrement`  will still get stuck at `6`

due to this logic in `AbsSeekBar`:

https://android.googlesource.com/platform/frameworks/base/+/711efaa0297f845c60a1de0acce680493f90bf02/core/java/android/widget/AbsSeekBar.java#590

Test Plan:
----------

### DIFF
--- a/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -216,6 +216,11 @@ public class ReactSlider extends AppCompatSeekBar {
       mStepCalculated = (mMaxValue - mMinValue) / (double) DEFAULT_TOTAL_STEPS;
     }
     setMax(getTotalSteps());
+
+    // setMax updates the internal "mKeyProgressIncrement" and on initial load the value will be set to
+    // 6 due to Math.round(DEFAULT_TOTAL_STEPS (128) / 20). And its hard to get it out of 6 if the range
+    // is not high enough so lets revert it back to 1 to respect mStep
+    setKeyProgressIncrement(1);
     updateLowerLimit();
     updateUpperLimit();
     updateValue();


package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -216,6 +216,11 @@ private void updateAll() {
       mStepCalculated = (mMaxValue - mMinValue) / (double) DEFAULT_TOTAL_STEPS;
     }
     setMax(getTotalSteps());
+
+    // setMax updates the internal "mKeyProgressIncrement" and on initial load the value will be set to
+    // 6 due to Math.round(DEFAULT_TOTAL_STEPS (128) / 20). And its hard to get it out of 6 if the range
+    // is not high enough so lets revert it back to 1 to respect mStep
+    setKeyProgressIncrement(1);
```
BartoszKlonowski, line 223:
```suggestion
    setKeyProgressIncrement(1);
```

BartoszKlonowski (approved):
This looks decent and works like a charm!
I would keep the explanation of the fix in the commit message or PR description (which is well written - thank you!) instead of putting it in the code, as after some time and multiple fixes it would get noisy with all the comments.

PR #578, opened 2024-02-23
Fix CI issues regarding deps

BartoszKlonowski:


BartoszKlonowski:
Looks like the problems with Windows platform are related to recent changes in depracation of `stdext::checked_array_iterator`.

PR #589, opened 2024-03-27
fix: new arch measurements on Android

okwasniewski:
Summary:
---------

This PR adds C++ shadow nodes to implement proper measuring for Android on Fabric/Bridgeless.

This code was based on Core implementation: https://github.com/facebook/react-native/blob/2af1da42ff517232f1309efed7565fe9ddbbac77/packages/react-native/ReactCommon/react/renderer/components/switch/androidswitch/react/renderer/components/androidswitch/AndroidSwitchShadowNode.cpp

### Before

CleanShot 2024-03-28 at 11 50 25@2x

### After

CleanShot 2024-03-28 at 11 51 00@2x

Test Plan:
----------

Check if on new architecture Slider is properly rendered without any styles

### DIFF
--- a/package/android/src/newarch/java/com/reactnativecommunity/slider/ReactSliderManager.java
+++ b/package/android/src/newarch/java/com/reactnativecommunity/slider/ReactSliderManager.java
@@ -1,11 +1,15 @@
 package com.reactnativecommunity.slider;
 
+import android.content.Context;
+import android.view.View;
 import android.widget.SeekBar;
 import androidx.annotation.Nullable;
 
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.common.MapBuilder;
+import com.facebook.react.uimanager.PixelUtil;
 import com.facebook.react.uimanager.SimpleViewManager;
 import com.facebook.react.uimanager.ThemedReactContext;
 import com.facebook.react.uimanager.UIManagerHelper;
@@ -17,6 +21,8 @@ import java.util.Map;
 import com.facebook.react.viewmanagers.RNCSliderManagerInterface;
 import com.facebook.react.viewmanagers.RNCSliderManagerDelegate;
 import com.facebook.react.module.annotations.ReactModule;
+import com.facebook.yoga.YogaMeasureMode;
+import com.facebook.yoga.YogaMeasureOutput;
 
 /**
  * Manages instances of {@code ReactSlider}.
@@ -190,11 +196,6 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> implement
     view.setOnSeekBarChangeListener(ON_CHANGE_LISTENER);
   }
 
-  @Override
-  public Map getExportedCustomDirectEventTypeConstants() {
-    return ReactSliderManagerImpl.getExportedCustomDirectEventTypeConstants();
-  }
-
   // these props are not available on Android, however we must override their setters
   @Override
   public void setMinimumTrackImage(ReactSlider view, @Nullable ReadableMap readableMap) {}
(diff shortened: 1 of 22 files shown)


package/react-native-slider.podspec
```diff
@@ -16,6 +16,13 @@ Pod::Spec.new do |s|
 
   s.source       = { :git => "https://github.com/callstack/react-native-slider.git", :tag => "v#{s.version}" }
   s.source_files = "ios/**/*.{h,m,mm}"
+
+  s.subspec "common" do |ss|
```
okwasniewski, line 20: [Question] Is there anyway we could use codegen as normal on iOS?

thymikee:
That's quite a lot of boilerplate to make it work. Is this expected in this case @cortinico?

package/android/src/newarch/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -210,4 +211,35 @@ public void setTapToSeek(ReactSlider view, boolean value) {}
...
+    @Override
+    public Map<String, Object> getExportedCustomBubblingEventTypeConstants() {
+        return ReactSliderManagerImpl.getExportedCustomBubblingEventTypeConstants();
+    }
+
+    @Nullable
+    @Override
+    public Map<String, Object> getExportedCustomDirectEventTypeConstants() {
+        return ReactSliderManagerImpl.getExportedCustomDirectEventTypeConstants();
+    }
```
okwasniewski, line 244: This now properly exports bubbling and direct events according to the JS spec

package/src/RNCSliderNativeComponent.ts
```diff
@@ -41,6 +41,6 @@ export interface NativeProps extends ViewProps {
   upperLimit?: Float;
 }
 
-export default codegenNativeComponent<NativeProps>(
-  'RNCSlider',
-) as HostComponent<NativeProps>;
+export default codegenNativeComponent<NativeProps>('RNCSlider', {
+  interfaceOnly: true,
```
cortinico, line 45: Q: What's the reason you don't use the generate C++ code and expand from it?

cortinico:
> That's quite a lot of boilerplate to make it work. Is this expected in this case @cortinico?

In theory you could use codegen and pick only the C++ files you need

package/src/RNCSliderNativeComponent.ts
okwasniewski: So when I removed this I got a new issue saying that I'm redefining ShadowNode with different types - how can I achieve what you mentioned to expand from generated C++ code? 

image

okwasniewski:
Hey, @BartoszKlonowski could you check if this fails locally for you on Android? For me it works fine but CI/CD fails for some reason

BartoszKlonowski:
> Hey, @BartoszKlonowski could you check if this fails locally for you on iOS? For me it works fine but CI/CD fails for some reason

The issue was that we first run the codegen but only then we pull the package with the current changes from the PR, so in this case the changes that required the codegen to apply were made **after** the codegen.
Changing the order of the steps so that we first pull the latest package from PR and then run codegen fixed it.

This was definitely an issue in general, but looks like previously we didn't have a chance to discover it. I'm glad we had that opportunity in this PR.

Thank you for notifying and pinging me! 👍

package/src/RNCSliderNativeComponent.ts
cortinico: Ok so it's a bit more complicated. Are you fine with merging this as it is, as it unblocks New Architecture migration, and then we can follow-up on a cleanup afterwards?

cortinico (approved):

okwasniewski:
Yup, I'm fine with merging this as is and clean it up later if we come up with better solutions in core. WDYT @thymikee @BartoszKlonowski ?

package/common/cpp/react/renderer/components/RNCSlider/RNCSliderState.cpp
```diff
@@ -0,0 +1,9 @@
+#include "RNCSliderState.h"
+
+namespace facebook {
```
BartoszKlonowski, line 3: We don't need that file if it's empty.
As long as the header is included anywhere with all the definitions then we should be fine while compilation. We only define source file with the definitions in it, if we want to compile the code once and link it anywhere when linking while building.

BartoszKlonowski (approved):
Let's have that, but we will need to create an issue to mark the cleanup.
I also left a couple of comments, but they can be addressed later too.

okwasniewski:
@BartoszKlonowski I've applied your review comments

PR #592, opened 2024-04-11
Fix missing header guard

alanjhughes:
Summary:
---------
When compiling for the new architecture the build fails on android because of a missing header guard in `RNCSliderMeasurementsManager.h`

Test Plan:
----------
Issue can be reproduced in the example app in the repo.  Adding the guard compiles successfully.

### DIFF
--- a/package/common/cpp/react/renderer/components/RNCSlider/RNCSliderMeasurementsManager.h
+++ b/package/common/cpp/react/renderer/components/RNCSlider/RNCSliderMeasurementsManager.h
@@ -1,4 +1,6 @@
 #ifdef ANDROID
+#pragma once
+
 #include <react/renderer/core/ConcreteComponentDescriptor.h>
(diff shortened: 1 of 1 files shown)


brentvatne (approved):

package/common/cpp/react/renderer/components/RNCSlider/RNCSliderMeasurementsManager.h
```diff
@@ -1,4 +1,6 @@
 #ifdef ANDROID
+#pragma once
```
BartoszKlonowski, line 2: Ah, yes! That's mine mistake - I got confused `#ifdef` with `ifndef` and suggested removing `once` during review of new arch measurement PR #589. 🙇‍♂️

BartoszKlonowski (approved):
🙏

PR #597, opened 2024-04-19
Fix: Linking during Android build for new arch

### LINKED ISSUE #594: Android build failure - C++ errors, buildCMakeDebug
## Environment
-My project won't build for Android with 4.5.1. 

Yes

React 18.2.0
React-Native 72.6

## Description

-Install slider package (latest version)

-iOS project works as expected

-Android project fails during build with errors related to:
C/C++: error: no matching constructor for initialization of 'facebook::react::PropsParserContext'
C/C++: error: redefinition of 'RNCSliderMeasurementsManager'
C/C++: error: non-virtual member function marked 'override' hides virtual member function
FAILURE: Build failed with an exception.
* What went wrong: Execution failed for task ':app:buildCMakeDebug[arm64-v8a]'.

  3 errors generated.
  ninja: build stopped: subcommand failed.
  
  C++ build system [build] failed while executing:
      /Users/joseph/Library/Android/sdk/cmake/3.22.1/bin/ninja \

BartoszKlonowski:
This pull request fixes #594 
It basically removes the unused and non-existing libraries from CMake config so that we don't require them when we don't really have them (nor build them on our own).

I briefly checked for the rest of the libs that are included there to see whether there's any other that can be removed, but it seems all other are indeed used and available during the build.

This pull request also introduces the job for building the new arch for Android, so that we can always check whether we are good there.

### DIFF
--- a/.github/workflows/ReactNativeSlider-CI.yml
+++ b/.github/workflows/ReactNativeSlider-CI.yml
@@ -131,7 +131,7 @@ jobs:
           cache-name: cached-gradle-new-arch-prop
         with:
           path: example/android/gradle.properties
-          key: ${{ hashFiles('./example/android/gradle.properties') }}
+          key: gradle-prop-new-arch-config
 
       - name: Build the Android OS app
         run: cd example/android && ./gradlew assembleDebug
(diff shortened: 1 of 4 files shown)


okwasniewski (approved):
Looks good! I wasn't sure if all of those libraries in CMakeLists are required. I guess some of them were introduced/exposed in 0.74.x (I was testing on RC).

artus9033 (approved):
Great, thanks for bringing the fix!

PR #598, opened 2024-04-25
added `value` prop to `StepMarker` component prop

biw:
Summary:
---------

Hello, thanks for the work on the library, it's great! I added a `value` prop to the recently added `StepMarker` prop. This allows developers to render StepMarker conditionally based on their value. This is a small change with a low service area but enables developers to implement features present in other slider libraries (sometimes called [`marks`](https://github.com/react-component/slider/))

My other solution, which would be larger, was making the `StepNumber` component something that could be passed in. Happy to update my PR in that direction if that would be helpful!

Test Plan:
----------

This is a small change, doesn't change any external APIs, and passes all `npm test`s. I have tested it in a web and native app but thought I'd put together a PR to share the changes upstream. Let me know how I can be more helpful on this front!

### DIFF
--- a/package/src/components/StepsIndicator.tsx
+++ b/package/src/components/StepsIndicator.tsx
@@ -44,6 +44,7 @@ export const StepsIndicator = ({
             <View style={styles.stepIndicatorElement}>
               <SliderTrackMark
                 key={`${index}-SliderTrackMark`}
+                value={i}
                 isTrue={currentValue === i}
                 thumbImage={thumbImage}
                 StepMarker={StepMarker}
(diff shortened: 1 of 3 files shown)


package/src/components/StepsIndicator.tsx
```diff
@@ -44,6 +44,7 @@ export const StepsIndicator = ({
             <View style={styles.stepIndicatorElement}>
               <SliderTrackMark
                 key={`${index}-SliderTrackMark`}
+                value={i}
```
BartoszKlonowski, line 47: `i` is the number of step and is constant for each. For having the current value on which the thumb is, `currentValue` should be used here.

BartoszKlonowski (commented):
Hello @biw, sorry for such delayed response.
This change you implemented is indeed required, but was already implemented in #581.
As for the docs update in README - let me handle that, as there will be some other things I will need to mention as well.

As this PR is no longer required, I will close it, but thank you very much for your time and for the contribution! 👍

PR #555, opened 2023-11-24
Feature: Custom steps marker

BartoszKlonowski:
This pull request closes #539 closes #240 closes #157 and fixes #514

It brings the custom steps indicator implementation to the Slider:

Android:

https://github.com/callstack/react-native-slider/assets/70535775/4940cb7f-faf3-486c-b5d4-eda0a94ff47b

iOS:

https://github.com/callstack/react-native-slider/assets/70535775/2c278f5a-2998-4e92-b63d-96762da75781

PR #603, opened 2024-05-21
Fix flicker issue on re-render in web implementation of Slider component

landabaso:
### Description

This PR addresses an issue with the `Slider` component where a flicker effect is visible during re-renders, specifically in the web implementation. The flicker was caused by the width being reset during the layout calculation. 

### Changes Made

- Modified `sliderStyle` to set the `width` only if it is not zero using the conditional spread operator.
  ```javascript
  const sliderStyle = { zIndex: 1, ...(width !== 0 && { width }) };


landabaso:
Oups, one of the tests failed (due to a snapshot comparison). Please let me know if you agree with this PR approach, and I'll update the snapshot accordingly.

landabaso:
In fact, I am still experiencing flicker issues on iOS and Android. Not setting the width at all (as it was originally) solves the flicker. That is: `const sliderStyle = { zIndex: 1 };` Is it strictly needed? However, I assume the width was set for a specific reason, which is why I didn't remove it in my initial approach.

Could you provide some insights on why the width was added, @BartoszKlonowski? This was introduced in [PR #555](https://github.com/callstack/react-native-slider/pull/555). Setting styles in `onLayout` can be tricky and may introduce flickering, depending on how parent components handle their children. 

Thanks!

Issue #617, opened 2024-07-10
Cannot read properties of undefined (reading 'toFixed')


rwilliamson1:
## Environment

- which version of react & react-native are you using?
- react: 18.3.1
- react-native: 0.74.2
- react-native-web: 0.19.10

## Description
When the slider is used with a step prop, when it is moved down to 0 it causes the error "Cannot read properties of undefined (reading 'toFixed')"
<br>
![Screenshot 2024-07-10 at 2 09 58 PM](https://github.com/callstack/react-native-slider/assets/169060610/03f8f34a-f41b-4d48-bc79-150f550b18fb)

## Reproducible Demo
The combination of using the step prop with the value going from a positive value down to 0 causes the error.
<br>
https://snack.expo.dev/jlLsyHnEs_ZzE3sODCTSB

tmizu23:
I'm experiencing the same issue. It was working fine up to v4.4.2. The problem started occurring after updating to v4.4.3.

events-jonas-chrisw:
having the same problem here...

events-jonas-chrisw:
is this fixed - I still get an error when setting the slider to 0.
I have had to revert to 4.4.2

PR #621, opened 2024-08-21
fix: set default initial value on web

draggie:
Summary:
---------
Referencing this issue: #617
React native web implementation had no default value set for initial value therefore resulting in such issues.
According to mobile implementation and documentation set the default value for initialValue prop to be 0.

Test Plan:
----------

Verified on controlled examples from example-web by setting initial useState value to undefined


BartoszKlonowski (approved):

events-jonas-chrisw:
is this fixed - I still get an error when setting the slider to 0.
I have had to revert to 4.4.2

Issue #596, opened 2024-04-18
 Unable to resolve module react-native/Libraries/NativeComponent/ViewConfigIgnore


raochenghui:
## Environment
Android
![image](https://github.com/callstack/react-native-slider/assets/70825629/582790ed-4a2b-47ce-905e-4a6b9cbd20c1)

* react-native info output:
```
System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 5800U with Radeon Graphics
    Memory: 1.04 GB / 14.83 GB
  Binaries:
    Node: 20.12.0 - 
    Yarn: 1.22.22 - 
    npm: 10.5.0 - 
    Watchman: 20240310.093207.0 - C:\ProgramData\chocolatey\bin\watchman.EXE
  SDKs:
    Android SDK: Not Found
    Windows SDK:
      AllowDevelopmentWithoutDevLicense: Enabled
      AllowAllTrustedApps: Enabled
      Versions: 10.0.22000.0
  IDEs:
    Android Studio: Not Found
    Visual Studio: Not Found
  Languages:
    Java: 17.0.10 
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2
    react-native: 0.68.2 => 0.68.2
    react-native-windows: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
```

* are you using the new architecture?

* which version of react & react-native are you using?
react-native@0.68.2

## Description
```bash
Error: Unable to resolve module react-native/Libraries/NativeComponent/ViewConfigIgnore from D:\project\APP\huangdao\hd_zhdd_app\node_modules\@react-native-community\slider\dist\RNCSliderNativeComponent.js: react-native/Libraries/NativeComponent/ViewConfigIgnore could not be found within the project or in these directories:
```
![image](https://github.com/callstack/react-native-slider/assets/70825629/36d05da6-8255-442f-953f-8a2f456679d1)

raochenghui:
react-native version 0.68.2 does not have this `ViewConfigIgnore.js`  file
![image](https://github.com/callstack/react-native-slider/assets/70825629/f3bd3313-8df9-405d-a826-f2380bc7cce0)

I switched the @react-native community/slider library to 4.5.0 to fix it

mirabo-trunghoang:
It is the same error when I tried to install with iOS, Xcode build log:
```
[...]

PR #626, opened 2024-08-26
chore: add info about current compatibility

draggie:
Summary:
---------
Regarding issue #596 
Recent upgrade of supported RN version removes support for versions older than RN 69, meaning that last compatible version of library is 4.5.0

Test Plan:
----------

No tests required


BartoszKlonowski (approved):
Alright! Thanks for checking that! 👍

Issue #571, opened 2024-01-30
Android: StackOverflow when `lowerLimit` is greater than `upperLimit` (and, on some devices, when the slider first loads)


personalizedrefrigerator:
## Environment

* react-native info output:
```
info Fetching system and libraries information...
System:
  OS: Linux 6.5 Ubuntu 23.10 23.10 (Mantic Minotaur)
  CPU: (16) x64 12th Gen Intel(R) Core(TM) i7-1260P
  Memory: 2.64 GB / 15.34 GB
  Shell:
    version: "5.9"
    path: /usr/bin/zsh
Binaries:
  Node:
    version: 18.13.0
    path: /usr/bin/node
  Yarn: Not Found
  npm: Not Found
  Watchman: Not Found
SDKs:
  Android SDK: Not Found
IDEs:
  Android Studio: Not Found
Languages:
  Java:
    version: 11.0.21
    path: /usr/bin/javac
  Ruby:
    version: 3.1.2
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.72.6
    wanted: 0.72.6
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: Not found
```

* are you using the new architecture? **No**

* which version of react & react-native are you using?
    * React Native **0.72.6**
    * React **18.2.0**

## Description

When `lowerLimit` is greater than `upperLimit`, `react-native-slider` can crash with a `StackOverflowException`.

> [!IMPORTANT]
>
> This is a reproducable version of [this downstream bug report](https://github.com/laurent22/joplin/issues/7974).
>
> The downstream bug report happens **while a slider is loading**, <u>even though `lowerLimit` and `upperLimit` are not set explicitly</u>. Rather, it happens because React Native sets props one-by-one, which can lead to a state where `lowerLimit > upperLimit` (and `realLowerLimit < realUpperLimit`, so `setProgress` works).
>
[...]

PR #638, opened 2024-09-10
fix: check for invalid configuration in limit settings

draggie:
Summary:
---------
fixes: #571 
Apply additional check and move the progress calculation to Slider class

Test Plan:
----------
Have ran the application examples on new/old arch witch success - the crash is not happening anymore.

### DIFF
--- a/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -246,7 +246,6 @@ public class ReactSlider extends AppCompatSeekBar {
   }
 
   /** Update limit based on props limit, max and min
-    * Fallback to lower limit if invalid configuration provided
   */
   private void updateUpperLimit() {
     double limit = Math.min(mRealUpperLimit, mMaxValue);
--- a/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
+++ b/package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
@@ -11,6 +11,7 @@ import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.drawable.BitmapDrawable;
 import android.os.Build;
+import android.util.Log;
 import android.util.AttributeSet;
 import android.view.accessibility.AccessibilityEvent;
 import android.view.accessibility.AccessibilityManager;
@@ -105,15 +106,13 @@ public class ReactSlider extends AppCompatSeekBar {
   }
 
   /*package*/ int getValidProgressValue(int progress) {
-      if (getLowerLimit() <= getUpperLimit()) {
-        if (progress < getLowerLimit()) {
-          progress = getLowerLimit();
-        } else if (progress > getUpperLimit()) {
-          progress = getUpperLimit();
-        }
+      if (progress < getLowerLimit()) {
+        progress = getLowerLimit();
+      } else if (progress > getUpperLimit()) {
+        progress = getUpperLimit();
       }
       return progress;
-    }
+  }
 
   /* package */ void setValue(double value) {
     mValue = value;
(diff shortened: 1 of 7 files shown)


package/ios/RNCSliderManager.m
```diff
@@ -158,8 +158,26 @@ - (void)sliderTouchEnd:(RNCSlider *)sender
...
+}
+RCT_CUSTOM_VIEW_PROPERTY(upperLimit, float, RNCSlider) {
+  float upperLimit = [RCTConvert float:json];
+
+  if (upperLimit < view.lowerLimit) {
+    upperLimit = view.lowerLimit;
+    NSLog(@"Invalid configuration: reverting lower limit to equal to upper limit")
+  }
+
+  view.upperLimit = upperLimit;
```
BartoszKlonowski, line 179:
```suggestion
  if (upperLimit < view.lowerLimit) {
    NSLog(@"Invalid configuration: upperLimit < lowerLimit; upperLimit not set")
  } else {
    view.upperLimit = upperLimit;
  }
```

package/ios/RNCSliderManager.m
```diff
@@ -158,8 +158,26 @@ - (void)sliderTouchEnd:(RNCSlider *)sender
...
-RCT_EXPORT_VIEW_PROPERTY(upperLimit, float);
+RCT_CUSTOM_VIEW_PROPERTY(lowerLimit, float, RNCSlider) {
+  float lowerLimit = [RCTConvert float:json];
+
+  if (lowerLimit > view.upperLimit) {
+    lowerLimit = view.upperLimit;
+    NSLog(@"Invalid configuration: reverting upper limit to equal to lower limit")
+  }
+
+  view.lowerLimit = lowerLimit;
```
BartoszKlonowski, line 169:
```suggestion
  if (upperLimit < view.lowerLimit) {
    NSLog(@"Invalid configuration: upperLimit < lowerLimit; lowerLimit not set")
  } else {
    view.lowerLimit = lowerLimit;
  }
```

package/src/Slider.tsx
```diff
@@ -281,6 +281,18 @@ const SliderComponent = (
...
+  const limitCheck = useCallback(() => {
+    if (lowerLimit >= upperLimit) {
+      console.warn(
+        'Invalid configuration: lower limit is supposed to me smaller than upper limit',
+      );
+    }
+  }, [lowerLimit, upperLimit]);
+
+  useEffect(() => {
+    limitCheck();
```
BartoszKlonowski, line 293: We don't need to create separate callback for `limitCheck` and use it only in the effect.
Let's have that single `if` check and `console.warn` directly in this effect.
Also, if the callback uses `lowerLimit, upperLimit` for dependencies then I don't see a reason for using anything different in the dependencies of this effect.

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -222,16 +232,28 @@ private void updateAll() {
...
-  /** Update limit based on props limit, max and min */
+  /** Update limit based on props limit, max and min
+   * Fallback to upper limit if invalid configuration provided
+  */
   private void updateLowerLimit() {
     double limit = Math.max(mRealLowerLimit, mMinValue);
-    mLowerLimit = (int) Math.round((limit - mMinValue) / (mMaxValue - mMinValue) * getTotalSteps());
+    int lowerLimit = (int) Math.round((limit - mMinValue) / (mMaxValue - mMinValue) * getTotalSteps());
+    if(lowerLimit > mUpperLimit) {
+      Log.w("Invalid configuration", "reverting lower limit to upper limit");
```
BartoszKlonowski, line 242: We shouldn't display logs in very different format for the same type of error.
Let's unify all of them to the one proposed in the suggestions for iOS platform (of course considering whether it's for `upperLimit` or `lowerLimit`).

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -222,16 +232,28 @@ private void updateAll() {
...
+    * Fallback to lower limit if invalid configuration provided
+  */
   private void updateUpperLimit() {
     double limit = Math.min(mRealUpperLimit, mMaxValue);
-    mUpperLimit = (int) Math.round((limit - mMinValue) / (mMaxValue - mMinValue) * getTotalSteps());
+    int upperLimit = (int) Math.round((limit - mMinValue) / (mMaxValue - mMinValue) * getTotalSteps());
+    if(mLowerLimit > upperLimit) {
+      Log.w("Invalid configuration", "reverting upper limit to equal to lower limit");
+    }
+    mUpperLimit = Math.max(upperLimit, mLowerLimit);
```
BartoszKlonowski, line 256:
```suggestion
    if (mLowerLimit > upperLimit) {
      Log.w("Invalid configuration", "upperLimit < lowerLimit; upperLimit not set");
    } else {
      mUpperLimit = upperLimit;
    }
```

BartoszKlonowski (commented):
If `lowerLimit > upperLimit` we shouldn't fallback to `upperLimit`. If we do that, then we will end up with having `lowerLimit === upperLimit`, which will prevent Slider from moving at all. This is not the best experience.
If a dev/user makes a simple mistake, typo, of having `lowerLimit` greater than the upper one, then we should just avoid setting that incorrect limit. In that worst case, the limit that is incorrect will just not work, not the whole Slider.

package/ios/RNCSliderComponentView.mm
```diff
@@ -272,6 +267,21 @@ - (void)loadImageFromImageSource:(ImageSource)source completionBlock:(RNCLoadIma
     }
 }
 
+void updateLimits(RNCSlider *slider, float newLowerLimit, float newUpperLimit) {
```
BartoszKlonowski, line 270: I believe this got skipped when applying changes, as we still fallback to the upper limit 🙃

package/src/Slider.tsx
```diff
@@ -281,6 +281,14 @@ const SliderComponent = (
           default: constants.LIMIT_MAX_VALUE,
         });
 
+  useEffect(() => {
+    if (lowerLimit >= upperLimit) {
+      console.warn(
+        'Invalid configuration: lower limit is supposed to me smaller than upper limit',
```
BartoszKlonowski, line 287:
```suggestion
        'Invalid configuration: lower limit is supposed to be smaller than upper limit',
```

package/android/src/main/java/com/reactnativecommunity/slider/ReactSlider.java
```diff
@@ -222,16 +232,30 @@ private void updateAll() {
...
+    if(lowerLimit > mUpperLimit) {
+      Log.d("Invalid configuration", "upperLimit < lowerLimit; lowerLimit not set");
+    }else {
+      mLowerLimit = Math.min(lowerLimit, mUpperLimit);
+    }
   }
 
-  /** Update limit based on props limit, max and min */
+  /** Update limit based on props limit, max and min
+    * Fallback to lower limit if invalid configuration provided
```
BartoszKlonowski, line 249: That's the caveat of comments in the code - they can quickly become outdated ;)
```suggestion
```

BartoszKlonowski (approved):
Looks very good! 👍

Issue #635, opened 2024-09-09
'react/renderer/components/rncore/EventEmitters.h' file not found


cjpete:
## Environment

* react-native info output:
```
System:
  OS: macOS 14.5
  CPU: (12) arm64 Apple M2 Pro
  Memory: 126.92 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 18.18.2
    path: ~/.nvm/versions/node/v18.18.2/bin/node
  Yarn:
    version: 3.6.4
    path: ~/.nvm/versions/node/v18.18.2/bin/yarn
  npm:
    version: 9.8.1
    path: ~/.nvm/versions/node/v18.18.2/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods: Not Found
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.0
      - iOS 17.0
      - macOS 14.0
      - tvOS 17.0
      - watchOS 10.0
  Android SDK:
    API Levels:
      - "30"
      - "31"
      - "33"
      - "34"
    Build Tools:
      - 30.0.3
      - 33.0.0
      - 33.0.1
      - 34.0.0
      - 35.0.0
    System Images:
      - android-34 | Google APIs ARM 64 v8a
      - android-35 | Google Play ARM 64 v8a
    Android NDK: Not Found
IDEs:
  Android Studio: 2024.1 AI-241.18034.62.2411.12071903
  Xcode:
    version: 15.0.1/15A507
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.12
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.3.1
    wanted: 18.3.1
  react-native:
    installed: 0.75.2
    wanted: 0.75.2
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

* are you using the new architecture? **Yes**

* which version of react & react-native are you using? 18.3.1 / 0.75.2 

## Description

When using the new architecture with statically linked frameworks (required if using react-native-firebase for example), the iOS app won't compile.

```
[...]

PR #639, opened 2024-09-11
Fix/ios new arch build

draggie:
Summary:
---------
fixes: #635 
This import is not necessary and caused the static frameworks build to fail while using new arch

Test Plan:
----------
Tested the application build on ios with new arch/old arch with/without static frameworks


BartoszKlonowski (approved):
👍

PR #655, opened 2024-10-08
Custom thumb's component with additional props

### LINKED ISSUE #154: Custom component for marker
Now it is only possible to pass a static image to the marker. It would be nice to support
custom components. For example to display some dynamic values inside of it.

BartoszKlonowski:
This pull request closes #154 
It implements additional possibilities of customizing the Slider's thumb component.
Including:
* minimum and maximum value are known at level of the thumb custom component,
* as well as the `currentValue` so that the main, custom component can know it's current position.
* It also adds the index of where that *custom step component* actually is and compare it to `currentValue` to know whether it is pass or before the thumb ("filled" or "empty").

---

What is still to be done before merging:
- [x] document all the additional props
Although it can be done in separate issue, as the docs would already use some attention to improve readability, especially considering extended options of customization.
- [x] Optimize
We currently render **a lot** each go, so it would be good to take a closer look
- [x] Analyze and fix the example case
The example implemented for new addition behaves incorrectly on iOS (the left section of custom thumb changes the position each step. It blinks for a moment before correctly render).

---

Below is the short demo with the implemented example:
(iOS to the left, Android on the right)

https://github.com/user-attachments/assets/b591ab24-fb41-4fd5-b80c-87c103b18171


dule-vikki:
@BartoszKlonowski  
I see that we just can add custom component for StepMarker only, right?
In this PR, Can we add custom component for Thumb?

BartoszKlonowski:
> @BartoszKlonowski I see that we just can add custom component for StepMarker only, right? In this PR, Can we add custom component for Thumb?

Yes, that is the goal of this PR. Although, need to mention that even with this PR it would still require using the `StepMarker` component, but to render it only if `stepMarked` is `true`. I didn't want to split this between step marker and thumb only, but I'm open for suggestions.

dule-vikki:
> > @BartoszKlonowski I see that we just can add custom component for StepMarker only, right? In this PR, Can we add custom component for Thumb?
> 
> Yes, that is the goal of this PR. Although, need to mention that even with this PR it would still require using the `StepMarker` component, but to render it only if `stepMarked` is `true`. I didn't want to split this between step marker and thumb only, but I'm open for suggestions.

@BartoszKlonowski 
OK. I'm looking forward to being able to custom Thumb component.

PR #668, opened 2024-11-15
Update deps and node modules

BartoszKlonowski:
This is basically a chore PR to update some deps and remove some. Worth to check though if everything runs smooth as usual and CI passes.


BartoszKlonowski:
I don't think this will be needed once #672 lands, let's close it.

PR #677, opened 2024-12-25
fix: ios inverted tap to seek in rtl

AymanHossam:
Summary:
---------

When `tapToSeek` option is enabled it works as expected on LTR, however after changing the application to RTL it seeks to the inverted tap location.

Test Plan:
----------

Before:

https://github.com/user-attachments/assets/61192081-7f05-4250-a052-2832dbb460da

After:

https://github.com/user-attachments/assets/832a03e6-e790-4c15-a5f4-c6a47946f8ec

### DIFF
--- a/package/ios/RNCSliderComponentView.mm
+++ b/package/ios/RNCSliderComponentView.mm
@@ -77,7 +77,14 @@ using namespace facebook::react;
 
     CGPoint touchPoint = [gesture locationInView:slider];
     float rangeWidth = slider.maximumValue - slider.minimumValue;
-    float sliderPercent = touchPoint.x / slider.bounds.size.width;
+    
+    float sliderPercent;
+    if ([UIView userInterfaceLayoutDirectionForSemanticContentAttribute:slider.semanticContentAttribute] == UIUserInterfaceLayoutDirectionRightToLeft) {
+        sliderPercent = 1.0 - (touchPoint.x / slider.bounds.size.width);
+    } else {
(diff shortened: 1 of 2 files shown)


ahmedawaad1804 (approved):

package/ios/RNCSliderComponentView.mm
```diff
@@ -77,7 +77,14 @@ - (void)tapHandler:(UITapGestureRecognizer *)gesture {
 
     CGPoint touchPoint = [gesture locationInView:slider];
     float rangeWidth = slider.maximumValue - slider.minimumValue;
-    float sliderPercent = touchPoint.x / slider.bounds.size.width;
+    
+    float sliderPercent;
+    if ([UIView userInterfaceLayoutDirectionForSemanticContentAttribute:slider.semanticContentAttribute] == UIUserInterfaceLayoutDirectionRightToLeft) {
```
BartoszKlonowski, line 82: Can't/Shouldn't we use `inverted` prop here?

BartoszKlonowski (approved):
In general this is good solution, thanks for taking care of that 👍 
I would really opt for using a common source of truth for the RTL before merging that.

package/ios/RNCSliderComponentView.mm
AymanHossam: I'm not passing the `inverted` prop, it's automatically gets inverted on the RTL layout. so in this case `inverted` would be `false`.

PR #682, opened 2025-01-27
Update RNCSliderComponentView.mm

ckknight:
Summary:
---------

This fixes a NSInvalidArgumentException which causes a full app crash when `<Slider accessibilityIncrements={['anything non-empty array']} />` is rendered, by initializing an NSMutableArray instead of an NSArray, as it called with `addObject:` if `accessibilityIncrements` is non-empty.

Test Plan:
----------

In iOS,

Render `<Slider minimumValue={1} maximumValue={3} accessibilityIncrements={['alfa', 'bravo', 'charlie']} />`

See if it crashes or not

----

Android should have no behavioral differences.


BartoszKlonowski (approved):
Thank you for this, @ckknight! 👍

PR #698, opened 2025-04-07
Move the project out of react-native-community, as it is supposed to

Sharcoux:
Summary:
---------

As explained [here](https://github.com/react-native-community/discussions-and-proposals/issues/330#issuecomment-772666380), the decision has been taken by the react-native-community that all projects should move out. This one is one of the latest that has not complied so far.

Test Plan:
----------

You'll need to:
- `npm publish` after this change to create the new entry on npm
- `npm deprecate @react-native-community/slider "This package has been moved to @callstack/slider and is no longer maintained. Please move to the new package for updates"`


BartoszKlonowski:
Welcome back, @Sharcoux.
Moving out of react-native-community is something I have in mind to do at some point, but not right now.
Changing this is something that requires some more communication and the proper moment when it comes to other changes around Slider and React Native itself, not only changing some files, unfortunately.
I will close this PR, but thank you for taking an interest in the matter.

PR #709, opened 2025-05-23
Release: v5.0.0  -  new architecture only

BartoszKlonowski:
This pull request releases Slider to version 5.0.0 with **new architecture only**.


package/package.json
draggie, line 1: I guess only thing is that maybe those dependencies should also be updated - like react from 18 to 19 and so on ?

BartoszKlonowski: All of them we have in devDependencies, it's up to us what version we use to create a lib, as long as it cooperates nicely with what user has as deps in their app.
We could update those, but I'd rather leave it for the further releases if possible.

draggie (approved):
Looks fine,

PR #714, opened 2025-06-05
Use `ImageURISource` for thumb image instead of `ImageSource` to fix image not loading

### LINKED ISSUE #711: Unable to add a thumbImage without having a StepMarker
## Environment

* react-native info output:
```
// react-native info
Expo v53
```

yes

React 19 and React Native v0.79.2

## Description

I am trying to put a thumbImage on my slider, however when I do so it don't work until I add a StepMarker. If I take off the StepMarker, the thumbImage will also be gone. I tried it on simulation iphone 16 with iOS 18.2 and with my iPhone 16 in real life also. I didn't find any examples with a Slider without StepMarker but with image.

## Reproducible Demo

Just create a Slider with a thumbImage and a StepMarker, then without StepMarker. The image should disappear and be replaced by default thumb.

```
<Slider
          minimumValue={0}
          maximumValue={100}
          step={5}
          value={value}
          onValueChange={handleChange}
          minimumTrackTintColor="#62A6E433"
          maximumTrackTintColor="#62A6E433"
          StepMarker={(): ReactElement => <View style={tw`h-5`}></View>}
          thumbTintColor={'#62A6E4'}
          thumbImage={require('@src/assets/images/SliderThumb.png') as ImageURISource}
        />
```

```
<Slider
          minimumValue={0}
          maximumValue={100}
          step={5}
          value={value}
          onValueChange={handleChange}
          minimumTrackTintColor="#62A6E433"
          maximumTrackTintColor="#62A6E433"
          // StepMarker={(): ReactElement => <View style={tw`h-5`}></View>}
          thumbTintColor={'#62A6E4'}
          thumbImage={require('@src/assets/images/SliderThumb.png') as ImageURISource}
        />
```

BartoszKlonowski:
This pull request fixes #711 

It migrates from `ImageSource` type from RN to `ImageURISource`. Those are both from RN and have common props however `ImageSource` is no longer supported by the v0.79 app (and beyond).

This fix includes both iOS and Android.
Also, all dev deps were updated to make sure we have the recent types used within the package when consumed by the app.

**Note:**: Removing tests from workflow is temporary due to versions update and tests now failing (some features of [React test](https://react.dev/warnings/react-test-renderer) were used previously).


BartoszKlonowski:
@brianlenz @Qdafengzi Thank you for the interest and checking the solution!
It's been a while since I had a possibility to sit down to it, but in the meantime, together with @okwasniewski (thanks!) we figured out that what is implemented in this PR is rather a workaround and we should rather stick to `ImageSource`. 
Considering input provided in #734 by @pingkai there is an option to figure out a better solution soon.
I'm hoping for your patience while we'll be pushing the topic forward, and in the meantime please reach out to a workaround suggested in #734.
I will keep this thread active as soon as there will be more findings or ETAs.
Thank you!

PR #722, opened 2025-08-06
fix: support for react native 0.81 on Android

gabrieldonadel:
Summary:
---------

Trying to use `react-native-slider` with react-native 0.81.0-rc.5 will cause an insta crash in the app due to an issue when parsing `RNCSliderProps` on Android

image

These changes are related to the Props 2.0 project and require us to set the necessary flags and definitions in Cmake, by using `target_compile_reactnative_options` this can be done automatically instead of setting them manually.
 

Test Plan:
----------

1. `npx @react-native-community/cli@latest init RC5 --skip-install --version next`
2. `yarn add @react-native-community/slider`
3. patch `package/android/src/main/jni/CMakeLists.txt`


BartoszKlonowski (approved):

okwasniewski:
@BartoszKlonowski can you release a new version with this fix? 🙏

PR #750, opened 2025-12-15
feat(ios): add swipeToSeek

sterlingwes:
Summary:
---------

I noticed on iOS that the slider does not behave as it does on Android when you make a swipe gesture on the track bar away from the thumb. On Android, the thumb snaps to the gesture point of the swipe and follows it along until the gesture ends. This feels more fluid and native, and allows for the user to be less precise in pressing the thumb but still find the precise value they want. `tapToSeek` doesn't allow for this since it expects a short press, but if a user is trying to grab the thumb and misses, nothing happens because the gesture is a swipe/pan.

I opted to conditionally add a pan gesture recognizer on iOS (which allows for swiping continuously in both directions) to achieve this based on an opt-in boolean prop `swipeToSeek`. This works fine alongside the existing tap recognizer for when tapToSeek is also enabled.

Test Plan:
----------

In the video below I'm demonstrating how it behaves relative to the tapToSeek behaviour with both steps defined and not. I also show that for sliders where the swipeToSeek prop is not set to true, it won't respond to those swipe gestures.

https://github.com/user-attachments/assets/ac021842-d1e6-4d49-a070-684b4e3c8ceb

I also verified the event handlers work the same by temporarily toggling the prop in these prop examples:

<details>
<summary>onSlidingStart / onSlidingComplete BEFORE (no swipeToSeek)</summary>

https://github.com/user-attachments/assets/daf719c1-b7ca-4fc7-abe6-ca2f840eeb65

</details>

<details>
<summary>onSlidingStart / onSlidingComplete AFTER (with swipeToSeek)</summary>

https://github.com/user-attachments/assets/2a252d3f-51d6-4e23-8ead-663b0a57949b

</details>


sterlingwes:
Seeking review from @BartoszKlonowski I think, whenever you have a moment? 🙏🏻🙇🏻‍♂️

sterlingwes:
hi @draggie  i notice you're in the middle of a rewrite of the library. Are you considering open issues like this in that or just working on parity with existing features?

BartoszKlonowski:
Hello @sterlingwes, thanks for reaching out!
Yes, in general, some time ago I decided to go ahead and rewrite the whole library to make it more up to date with standards (Kotlin, Swift) and to migrate it to new UI also be able to get rid of native limitation current Slider suffers.
Because of that I didn't yet merge all the pending PRs that are, of course, targeting the "old" Slider.
Still, this PR you created will be merged and released along with some other features (like two thumbs by @draggie) in near future, as part of yet Slider v5 (before v6 rewrite kicks in) but because of limited time I have, the ETA for that is still to be figured out.
I totally appreciate all the contributions, including yours 💪  but at the same time I have no choice but to count on your patience 🙏

sterlingwes:
awesome thanks @BartoszKlonowski for the update 🙏🏻  sounds good

package/ios/RNCSliderComponentView.mm
```diff
@@ -24,6 +24,7 @@ @implementation RNCSliderComponentView
     RNCSlider *slider;
     UIImage *_image;
     BOOL _isSliding;
+    BOOL _swipeGestureEnabled;
```
BartoszKlonowski: @sterlingwes Can you elaborate on why we need this private variable?
From what I see it's only used once in the whole runtime to set the gesture recognizer. But is this required?
If we could make the `swipeToSeek` being `true` by default, we wouldn't need this additional var check.

package/ios/RNCSliderComponentView.mm
```diff
@@ -298,6 +302,78 @@ - (void)setInverted:(BOOL)inverted
...
+        [slider addGestureRecognizer:panGesturer];
+        _swipeGestureEnabled = YES;
+    }
+}
+
+- (void)panHandler:(UIPanGestureRecognizer *)gesture {
+    CGPoint location = [gesture locationInView:slider];
+    
+    switch (gesture.state) {
+
```
BartoszKlonowski:
```suggestion
```

package/ios/RNCSliderComponentView.mm
```diff
@@ -298,6 +302,78 @@ - (void)setInverted:(BOOL)inverted
...
+        ->onRNCSliderValueChange(RNCSliderEventEmitter::OnRNCSliderValueChange{.value = static_cast<Float>(slider.value)});
+    }
+    
+    slider.lastValue = discreteValue;
+}
+
+- (float)calculateSliderValueFromLocation:(CGPoint)point {
+    CGFloat sliderWidth = slider.bounds.size.width;
+    
+    if (sliderWidth <= 0) {
```
BartoszKlonowski: When such case is possible?

package/ios/RNCSliderComponentView.mm
```diff
@@ -298,6 +302,78 @@ - (void)setInverted:(BOOL)inverted
...
+        }
+            
+        case UIGestureRecognizerStateChanged: {
+            [self updateSliderToLocation:location];
+            break;
+        }
+            
+        case UIGestureRecognizerStateEnded:
+        case UIGestureRecognizerStateCancelled:
+        case UIGestureRecognizerStateFailed: {
```
BartoszKlonowski: Don't we want to avoid sending an event if the state is failed?
What will be the `value` sent with this event? Will it still be updated despite being failed, or will we send duplicated position?
Let's handle the `UIGestureRecognizerStateFailed` separately, I'm fine having it doing nothing, unless you have some other ideas.

BartoszKlonowski (changes requested):
I have some simple comments/questions, but I "request changes" because I think that the `swipeToSeek` should be hardcoded to `true`.
The reason is that in previous versions of iOS platform this is how Slider was working, and there was no option to turn off the behavior of being able to swipe from any position. I would like to keep it that way, instead of introducing new prop that will be instantly turned on by (I assume) majority of consumers.
@sterlingwes Are you OK adjusting to that idea or do you have a strong reasoning behind the current approach?
Let me know :)

package/ios/RNCSliderComponentView.mm
sterlingwes: totally open to making this the default behaviour and not needing the tracking bool 👍🏻 

the opt-in prop was intended to allow this as a minor release but if you're planning on a major and open to the behavioural change, makes sense to me to simplify for sure

package/ios/RNCSliderComponentView.mm
sterlingwes: i interpreted `onSlidingStart` and `onSlidingComplete` as being a callback pair where the assumption is that an event for the first would eventually be followed by an event for the second. If we didn't emit an event for the failed case it'd be "dangling". If that's the wrong interpretation of how these could be used I think the no-op makes sense 👍🏻

package/ios/RNCSliderComponentView.mm
sterlingwes: good catch! seems like an overkill guard given this path only gets invoked on slider changes

package/ios/RNCSliderComponentView.mm
BartoszKlonowski: Hold on, a major? I will need to double check if this behavior was there in the old Sliders, or if my memory if faulty here 🤔
Thanks for the comment!

BartoszKlonowski: Alright, I checked, and this indeed wasn't the behavior back then.
@sterlingwes In fact, and why I was confused at first, the goal was always to have all platforms behaving in the same manner. So I consider lack of this `swipeToSeek` feature as a bug, which your PR will "fix" by aligning iOS platform to Android.
So I say: no worries, having this as the default behavior can still be fit into near minor release 👍 

Let me know if you need any help converting this implementation into the "hardcoded" version :)

sterlingwes:
Updated @BartoszKlonowski per your feedback, sorry for the delay. This simplifies things quite a bit, thanks! ✨ 

Proof it still works w/ rebuild:


https://github.com/user-attachments/assets/a3fdfa09-2559-4372-b2a9-539169126c3f

PR #751, opened 2025-12-18
fix(web): align step markers with thumb position using thumbSize prop

### LINKED ISSUE #743: Web: Step markers don't align with thumb positions
## Environment

* are you using the new architecture? yes

```
    "react": "19.1.0",
    "react-native": "0.81.4",
```

## Description

On Web, stepNumber and StepMarker don't align with the thumb position towards the left and right edge. For the center value it matches, but the further you go left and right the further off it is.

Image

Image

Image

When comparing Web and Android, the underlying problem might also be that on web, the whole slider has less margin left and right. 

On web, the slider hast the same width as the Button below:

Image

Same code on Android, the slider has less width and the step markers align:

Image

## Reproducible Demo

https://snack.expo.dev/@bananer_rw/react-native-slider-step-marker-repro

mikolajwilczek:
Fixes #743

Summary:
---------

- Fix step marker alignment on web platform edges
- Pass thumbSize prop to StepsIndicator component
- Use actual thumbSize value instead of hardcoded constant
- Add WebProps type for proper TypeScript typing

Test Plan:
----------
- Verified the change only affects the web platform - native platforms (iOS/Android) remain unchanged
- Added web example with step numbers to demonstrate the fix
- Tested step marker alignment at left edge, center, and right edge positions - all markers now align correctly with thumb positions
- Tested with default and custom `thumbSize` values on web - alignment works correctly when `thumbSize` is smaller than the gap between steps
- Regression tested on iOS, Android, and web (Safari and Chrome) - all existing examples continue to work
Screenshot 2025-12-18 at 20 12 13

### DIFF
--- a/package/src/components/StepsIndicator.tsx
+++ b/package/src/components/StepsIndicator.tsx
@@ -23,7 +23,6 @@ export const StepsIndicator = ({
   renderStepNumber?: boolean;
   thumbImage?: ImageSource;
   isLTR?: boolean;
-  thumbSize?: number;
 }) => {
   const stepNumberFontStyle = useMemo(() => {
     return {


README.md
```diff
@@ -96,8 +96,9 @@ To use this library you need to ensure you are using the correct version of Reac
 | `maximumTrackImage` | Assigns a maximum track image. Only static images are supported. The leftmost pixel of the image will be stretched to fill the track. | Image<br/>.propTypes<br/>.source | iOS |
 | `minimumTrackImage` | Assigns a minimum track image. Only static images are supported. The rightmost pixel of the image will be stretched to fill the track. | Image<br/>.propTypes<br/>.source | iOS |
 | `thumbImage` | Sets an image for the thumb. Only static images are supported. Needs to be a URI of a local or network image; base64-encoded SVG is not supported. | Image<br/>.propTypes<br/>.source | |
+| `thumbSize` | Define the size of the thumb. Only for web<br/>Default value is 20px. | number | web |
```
BartoszKlonowski: We don't want to add this as a new prop - doing this will bring **a lot** of confusion for users.
Devs been asking for such customization for a long time now, expecting a possibility to actually increase the **visible size** of the thumb. What we have here, instead, is the "under the hood" layout width of the thumb that we reserve for the proper alignment when rendering Slider with steps number.
Let's remove this `thumbSize` prop from where relevant (README, Slider.tsx) and let's make it a const just for the implementation detail purpose.

mikolajwilczek:
Please let me know if anything needs to be aligned now :)

package/src/components/StepsIndicator.tsx
```diff
@@ -15,6 +15,7 @@ export const StepsIndicator = ({
   renderStepNumber,
   thumbImage,
   isLTR,
+  thumbSize = constants.THUMB_SIZE,
```
BartoszKlonowski:
```suggestion
```

BartoszKlonowski:
OK, for some reason GitHub kept on ignoring removal of lines when trying to apply suggestion in a batch, so forgive me for adding so many commits on top of your branch, @mikolajwilczek.
I'll merge this as soon as we have checks passing.

PR #752, opened 2025-12-29
feat: add thumbSize to set size of thumb

bdtren:
Summary:
---------

This PR adds a new thumbSize prop to control the thumb’s size for both with and without thumb image. It is a simple feature that I need and did patch-package in some of my projects but I didn't have a chance to contribute, I think some developers might also need it:
| Android | iOS |
|---|---|
| Android | iOS |

Test Plan:
----------

* Install/refresh package:
```
npm install
npm run refresh-package
```

* Android:
```
npm run example-android
```

Verify in the example list:
“Custom thumb size (no image)” renders a larger thumb (e.g. thumbSize={32}) tinted with thumbTintColor.
“Custom thumb size (scaled image)” renders the thumb image scaled to the configured size (e.g. thumbSize={60}).

* iOS:
```
npm run example-ios
```
Verify the same two examples behave equivalently on iOS (thumb resizes; image scales when provided).

### DIFF
--- a/package/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
+++ b/package/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
@@ -134,7 +134,7 @@ public class ReactSliderManager extends SimpleViewManager<ReactSlider> implement
 
   @Override
   @ReactProp(name = "thumbSize", defaultFloat = 0f)
-  public void setThumbSize(ReactSlider view, double size) {
+  public void setThumbSize(ReactSlider view, float size) {
     ReactSliderManagerImpl.setThumbSize(view, size);
   }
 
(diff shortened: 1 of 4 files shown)


BartoszKlonowski:
Hello @bdtren, let me know please if you plan to push this forward, or should I take it from here?

package/android/src/main/java/com/reactnativecommunity/slider/ReactSliderManager.java
```diff
@@ -132,6 +132,12 @@ public void setThumbTintColor(ReactSlider view, Integer color) {
     ReactSliderManagerImpl.setThumbTintColor(view, color);
   }
 
+  @Override
+  @ReactProp(name = "thumbSize", defaultFloat = 0f)
+  public void setThumbSize(ReactSlider view, double size) {
```
BartoszKlonowski:
```suggestion
  public void setThumbSize(ReactSlider view, float size) {
```

PR #821, opened 2026-07-17
fix(ios): load thumbImage/trackImage without the legacy bridge on bridgeless (new arch)

### LINKED ISSUE #773: thumbImage not working on Rn 0.84.0
## Environment

* react-native info output:
System:
  OS: macOS 26.3
  CPU: (14) arm64 Apple M4 Pro
  Memory: 313.73 MB / 24.00 GB
  Shell:
    version: 3.2.57
    path: /bin/bash
Binaries:
  Node:
    version: 22.19.0
    path: /Users/jibeleck/.nvm/versions/node/v22.19.0/bin/node
  Yarn: Not Found
  npm:
    version: 10.9.3
    path: /Users/jibeleck/.nvm/versions/node/v22.19.0/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.15.2
    path: /Users/jibeleck/.rbenv/shims/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2025.2 AI-252.27397.103.2522.14514259
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.8
    path: /usr/bin/javac
  Ruby:
    version: 2.7.4
    path: /Users/jibeleck/.rbenv/shims/ruby
npmPackages:
  "@react-native-community/cli":
    installed: 20.1.1
    wanted: 20.1.1
  react:
    installed: 19.2.3
    wanted: 19.2.3
  react-native:
    installed: 0.84.0
    wanted: 0.84.0
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: true
iOS:
  hermesEnabled: true
  newArchEnabled: true

yes

    "@react-native-community/slider": "5.1.2",
    "react-native": "0.84.0",

## Description
custom thumbImage stoped working after update to latest RN (not sure if problem is latest RN or enabling new arch).

Problem seem to be in RNCSliderComponentView.mm - loadImageFromImageSource is relying on bridge.  

I am appending patch package generated by Cloude LLM to bypass this issue.

[@react-native-community+slider+5.1.2.patch](https://github.com/user-attachments/files/25500669/%40react-native-community%2Bslider%2B5.1.2.patch)

arron-kukadia:
## Summary

On bridgeless React Native (the default new-architecture mode since RN 0.74+), `thumbImage`, `trackImage`, `minimumTrackImage` and `maximumTrackImage` silently fail to render on iOS. `loadImageFromImageSource` resolves the ImageLoader through `[RCTBridge currentBridge]` (flagged by the existing TODO in the file), which doesn't reliably provide a working ImageLoader without the legacy bridge — the completion never delivers an image, and the slider falls back to the default thumb.

Fixes #773

## Fix

Keep the existing ImageLoader path as the first attempt (no behaviour change where it works). If the module is unavailable **or** returns no image, fall back to fetching the image directly with `NSURLSession` using the same `NSURLRequest` — this covers bundled `file://` assets, Metro URLs in dev, and remote images, and honours custom request headers. The retry-on-no-image (rather than only on missing module) is deliberate: on bridgeless, the interop layer can hand back an ImageLoader instance that exists but cannot load, so gating on module presence alone would not fix the bug.

Details:
- The `NSURLSession` completion invokes the callbacks on the main queue, since `failureBlock` call sites in `updateProps` mutate UIKit directly and previously only ran synchronously on the main thread.
- The scale fallback uses `RCTScreenScale()` (not `[UIScreen mainScreen]`, which is unavailable on visionOS — this pod declares visionOS support).
- The stale TODO comment is replaced with one describing the current dual-path behaviour.

Refactored `loadImageFromImageSource` uses legacy-bridge path first: when the bridge is present, RCTImageLoader provides caching, decode-time sizing and custom RCTImageURLLoader schemes.
On bridgeless, [RCTBridge currentBridge] may be nil or expose an ImageLoader that cannot load
(https://github.com/reactwg/react-native-new-architecture/discussions/31#discussioncomment-2717047),
so fall back to fetching the bytes directly with NSURLSession.
The fallback covers file:// (bundled require() assets) and http(s) (Metro in dev, remote images) and honours custom request headers;
it does not support RCTImageURLLoader plugin schemes (e.g. ph://) or ImageLoader's
resizing and caching. TODO: replace both paths with the Fabric ImageManager pipeline.

## Known limitations of the fallback

These apply only on bridgeless, where the fallback is effectively the primary path, and match the documented `thumbImage` API surface (static local/network images):

- `RCTImageURLLoader` plugin schemes (e.g. `ph://`, custom app-registered schemes) are not supported by `NSURLSession` and will fail; they were undocumented for this component but worked via the bridge ImageLoader on the old architecture.
- The fallback does not apply ImageLoader's decode-time resizing (`source.size` + cover mode) or decoded-image caching. A remote `{uri, width, height}` source without `thumbSize` decodes and renders at intrinsic bitmap size.

Both could be addressed by resolving `RCTImageLoader` from the new-arch module registry (or ultimately the Fabric ImageManager pipeline) — left out of scope here to keep this a minimal fix for #773.

## Test plan

Verified in a production Expo SDK 56 / RN 0.85 app (bridgeless, new arch): before the change `thumbImage` renders the default circle; after it, the custom thumb renders correctly in both dev (Metro-served asset) and release (bundled asset). Android untouched.

### DIFF
--- a/package/ios/RNCSliderComponentView.mm
+++ b/package/ios/RNCSliderComponentView.mm
@@ -274,14 +274,6 @@ using namespace facebook::react;
 }
 
 
-// Legacy-bridge path first: when the bridge is present, RCTImageLoader provides caching,
-// decode-time sizing and custom RCTImageURLLoader schemes. On bridgeless, [RCTBridge currentBridge]
-// may be nil or expose an ImageLoader that cannot load
-// (https://github.com/reactwg/react-native-new-architecture/discussions/31#discussioncomment-2717047),
-// so fall back to fetching the bytes directly with NSURLSession. The fallback covers file://
-// (bundled require() assets) and http(s) (Metro in dev, remote images) and honours custom request
-// headers; it does not support RCTImageURLLoader plugin schemes (e.g. ph://) or ImageLoader's
-// resizing and caching. TODO: replace both paths with the Fabric ImageManager pipeline.
 - (void)loadImageFromImageSource:(ImageSource)source completionBlock:(RNCLoadImageCompletionBlock)completionBlock failureBlock:(RNCLoadImageFailureBlock)failureBlock
 {
     NSString *uri = [[NSString alloc] initWithUTF8String:source.uri.c_str()];
--- a/package/ios/RNCSliderComponentView.mm
+++ b/package/ios/RNCSliderComponentView.mm
@@ -9,6 +9,7 @@
 #import <React/RCTBridge+Private.h>
 #import "RCTImagePrimitivesConversions.h"
 #import <React/RCTImageLoaderProtocol.h>
+#import <React/RCTUtils.h>
 #import "RCTFabricComponentsPlugins.h"
 #import "RNCSlider.h"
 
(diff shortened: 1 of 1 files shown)


package/ios/RNCSliderComponentView.mm
```diff
@@ -273,23 +274,61 @@ - (void)updateProps:(const Props::Shared &)props oldProps:(const Props::Shared &
...
 
-// TODO temporarily using bridge, workaround for https://github.com/reactwg/react-native-new-architecture/discussions/31#discussioncomment-2717047, rewrite when Meta comes with a solution.
+// Legacy-bridge path first: when the bridge is present, RCTImageLoader provides caching,
+// decode-time sizing and custom RCTImageURLLoader schemes. On bridgeless, [RCTBridge currentBridge]
+// may be nil or expose an ImageLoader that cannot load
+// (https://github.com/reactwg/react-native-new-architecture/discussions/31#discussioncomment-2717047),
+// so fall back to fetching the bytes directly with NSURLSession. The fallback covers file://
+// (bundled require() assets) and http(s) (Metro in dev, remote images) and honours custom request
+// headers; it does not support RCTImageURLLoader plugin schemes (e.g. ph://) or ImageLoader's
+// resizing and caching. TODO: replace both paths with the Fabric ImageManager pipeline.
```
BartoszKlonowski: Let's move it either to the commit message or to the PR's description. 🙏 

```suggestion
```

BartoszKlonowski (approved):
It works like a charm and I couldn't find anything wrong with it, great job! 👍

